Issue #1, opened 2016-01-22
Pull in Presence CRDT into project


chrismccord:
At your earliest convenience @asonge . I can even handle the namespace rename if you just want to copy the impl and test files over. I just want want to make sure your contribution stays intact :)

PR #2, opened 2016-01-22
Refactor and import of presence CRDTs

asonge:


asonge:
Fixes issue #1

lib/phoenix_pubsub/tracker/exceptions.ex
```diff
@@ -0,0 +1,3 @@
+defmodule Phoenix.Tracker.State.Invariant do
```
josevalim: `InvariantError`?

test/phoenix_pubsub/tracker_state_test.exs
```diff
@@ -0,0 +1,116 @@
+defmodule Phoenix.TrackerStateTest do
+  use ExUnit.Case
```
josevalim: async: true?

test/phoenix_pubsub/tracker_state_test.exs
```diff
@@ -0,0 +1,116 @@
...
+  # New print-callback Presence
+  defp newp(node) do
+    TrackerState.new({node, 1})
+  end
+
+  defp new_conn() do
+    make_ref()
+  end
+
+  test "That this is set up correctly" do
```
josevalim: test names should start in lower case :)

lib/phoenix_pubsub/tracker/state.ex
```diff
@@ -0,0 +1,290 @@
+defmodule Phoenix.Tracker.State do
```
josevalim: We want to `@moduledoc false` pretty much all of those!

test/phoenix_pubsub/tracker_state_test.exs
chrismccord: We'll have to revisit this, but I made it false because we are spawning trackers on nodes. If we keep the registered tracker names and topics unique it may be okay however

josevalim:
This is fantastic @asonge. :heart: I am sorry I cannot contribute to the discussion with more than cosmetic stuff.

test/phoenix_pubsub/tracker_state_test.exs
josevalim: Cool, ignore me then :D

chrismccord:
<3 <3 <3

asonge:
Thanks, there's still a lot of inter-leavings and untested API functions. I'll probably be tackling code coverage on this over the weekend.

PR #7, opened 2016-01-27
Pull in PubSub

chrismccord:


lib/phoenix_pubsub/local.ex
```diff
@@ -0,0 +1,230 @@
...
+
+  @doc """
+  Sends a message to all subscribers of a topic.
+
+    * `pubsub_server` - The registered server name or pid
+    * `topic` - The string topic, for example "users:123"
+
+  ## Examples
+
+      iex> broadcast(:pubsub_server, self, "foo")
```
josevalim: Sorry, we need to update the docs.

josevalim: Actually, not sorry. All of these docs are outdated from the time we added shards. :)

PR #24, opened 2016-03-07
Fix monitor leak

chrismccord:
@josevalim I imagine you'll want to chat about this one. The subscriber is responsible for calling :subscribe/:unsubscribe only if they have a new topic to subscribe to or an existing one to unsubscribe from. For 1M subscribers on my laptop, subscription lookup time is ~ 50 microsecs, so it's not too costly and the lookup happens in the caller's context. The only work added to the local server is keeping tracking of the monitor refs and subscription counts, but it should be minimal. Thoughts?

### DIFF
--- a/lib/phoenix_pubsub/local.ex
+++ b/lib/phoenix_pubsub/local.ex
@@ -36,13 +36,14 @@ defmodule Phoenix.PubSub.Local do
 
   """
   def subscribe(pubsub_server, pool_size, pid, topic, opts \\ []) when is_atom(pubsub_server) do
-    {:ok, {topics, pids}} =
-      pubsub_server
-      |> local_for_pid(pid, pool_size)
-      |> GenServer.call({:subscribe, pid, topic, opts})
+    {local, gc} =
+      pid
+      |> pid_to_shard(pool_size)
+      |> pools_for_shard(pubsub_server)
 
-    true = :ets.insert(topics, {topic, {pid, opts[:fastlane]}})
-    true = :ets.insert(pids, {pid, topic})
+    :ok = GenServer.call(local, {:subscribe, pid, opts})
+    true = :ets.insert(gc, {pid, topic})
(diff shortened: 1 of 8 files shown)


chrismccord:
@josevalim I fixed the races here by moving the operations on the pids table inside the local server, and using select_delete as an optimized monitor count. For 1M subscribers, inserts are usually < 10 microsecs, deletes are < 20 microsecs. I will continue spiking the ref API, but I'm leaning more and more towards the approach in this PR.

chrismccord:
@josevalim ready for review

lib/phoenix_pubsub/local.ex
```diff
@@ -186,30 +189,36 @@ defmodule Phoenix.PubSub.Local do
...
-  def handle_call({:subscribe, pid, _topic, opts}, _from, state) do
+  def handle_call({:subscribe, pid, opts}, _from, state) do
     if opts[:link], do: Process.link(pid)
-    Process.monitor(pid)
-    {:reply, {:ok, {state.topics, state.pids}}, state}
+    {:reply, :ok, put_new_monitor(state, pid)}
+  end
+
+  def handle_call({:unsubscribe, pid, topic}, _from, %{gc: gc} = state) do
+    :ets.match_delete(gc, {pid, topic})
```
chrismccord: I think we have a race here. We'll need to move {pid, topic} insert into the subscribe

chrismccord: We could get around the race by moving back to the `{ref, count}` state and using `:ets.select_delete` as the atomic op on unsubscribe. We could still offload the ets insert to the client in that case so the only cost is tracking the ref count. Thoughts?

chrismccord: There's a race in that case as well. I think we have no choice but to move pids insert into the server

josevalim:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #25, opened 2016-03-12
Clean up warnings emissions during tests.

mwmiller:
- Stop using the PID-version of subscribe/unsubscribe where possible
- Expose undocumented `call` for the remaining case
- Add explicit test for the deprecation warnings

I tried to fix the `h` documentation for `subscribe` and `unsubscribe`
to only show the non-deprecated versions, but failed in the `subscribe`
case. I think `unsubscribe` might have worked because of the differing
arities between recommended and deprecated. Rather than have it
half-done I'll leave it to someone more familiar.

Please let me know if any of these changes are worrisome.  I intend to work on a more in-depth feature in the future, but wanted to see a "clean" test run before starting on anything extensive.

### DIFF
--- a/lib/phoenix_pubsub/pubsub.ex
+++ b/lib/phoenix_pubsub/pubsub.ex
@@ -283,7 +283,9 @@ defmodule Phoenix.PubSub do
     call(server, :node_name, [])
   end
 
-  defp call(server, kind, args) do
+  @doc false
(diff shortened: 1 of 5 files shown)


chrismccord:
I have these cleaned up locally, but I need to rework the test helpers that simulate net splits. I appreciate the help though!

PR #26, opened 2016-03-16
Add ets based ORSWOT

chrismccord:
@josevalim @asonge I still need to add specs but I've added an ets-based set and performance seems reasonable for an initial release. Merging two 100k element sets takes 3.5s. Looking up members by topic for only online replicas takes 35ms. There are still many optimizations we can do, but the next steps are as follows:

- verify the implementation is correct
- expand our test cases
- add generational deltas

Let me know how it looks!


lib/phoenix_pubsub/tracker.ex
```diff
@@ -187,7 +187,7 @@ defmodule Phoenix.Tracker do
     server_name
     |> GenServer.call({:list, topic})
     |> State.get_by_topic(topic)
-    |> Enum.map(fn {_conn, key, meta} -> {key, meta} end)
+    |> Enum.map(fn {{_pid, _topci}, {{key, meta}, _}} -> {key, meta} end)
```
Gazler: _topci -> _topic

lib/phoenix_pubsub/tracker/state.ex
```diff
@@ -1,291 +1,283 @@
...
+    end
+
+    {delta, normalized_list}
+  end
 
   @doc """
-  Joins any 2 dots together.
+  Merges two sets, or a delta into a set.
+
+  Returns a 3-tuple of the updated set, and the joiend and left elements.
```
Gazler: joiend -> joined

test/phoenix_pubsub/tracker_state_test.exs
```diff
@@ -1,68 +1,74 @@
-defmodule Phoenix.TrackerStateTest do
+defmodule Phoenix.StateTest do
   use ExUnit.Case
 
-  alias Phoenix.Tracker.State, as: TrackerState
+  alias Phoenix.Tracker.State, as: State
```
Gazler: You can just do `alias Phoenix.Tracker.State` now.

Gazler:
@chrismccord I don't have a deep enough understanding (yet!) to provide anything more valuable than those comments. The docs and tests made sense to me though.

chrismccord:
@Gazler thanks for the feedback as usual!

PR #30, opened 2016-04-05
Replace occurances of "master/slave" terminology with "primary/replica"

folz:
The code, docs, and some tests contain references to a master/slave replication configuration.
While this terminology has been used for a long time, those terms may carry racially charged meanings to users.
This patch replaces all occurrences of master and slave with 'primary' and 'replica', following the Django Project's lead.

@see: https://github.com/django/django/pull/2692 and https://github.com/django/django/pull/2694

N.B. - there are two private functions in `tests/` where the slave terminology is still used, because they are calling erlang directly. This should be the exception and not the rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
 
 ## Testing
 
-Testing by default spawns slave nodes internally for distributed tests.
-To run tests that do not require slave nodes, exclude  the `clustered` tag:
+Testing by default spawns replica nodes internally for distributed tests.
+To run tests that do not require replica nodes, exclude  the `clustered` tag:
 
     $ mix test --exclude clustered
--- a/config/test.exs
+++ b/config/test.exs
@@ -2,7 +2,7 @@ use Mix.Config
 
 config :phoenix_pubsub,
   pubsub: [Phoenix.PubSub.Test.PubSub, [pool_size: 1]],
-  nodes: [:"slave1@127.0.0.1", :"slave2@127.0.0.1"]
+  nodes: [:"replica1@127.0.0.1", :"replica2@127.0.0.1"]
 
 config :logger,
   level: :info,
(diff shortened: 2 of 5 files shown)


lib/phoenix_pubsub/pubsub.ex
```diff
@@ -81,20 +81,20 @@ defmodule Phoenix.PubSub do
 
     * `subscribe` - subscribes the given pid to the given topic
       sends:        `{:subscribe, pid, topic, opts}`
-      respond with: `:ok | {:error, reason} | {:perform, {m, f, a}}`
+      respond with: `:ok | {:error, reason} | {:perform, {p, f, a}}`
```
josevalim: The `m` here is for `module` and not `master`. Please revert the changes to this file. :)

lib/mix/tasks/phoenix.pub_sub.bench.ex
```diff
@@ -17,36 +17,36 @@ defmodule Mix.Tasks.Phoenix.PubSub.Bench do
     size       = opts[:size] || 100_000
     delta_size = opts[:delta_size] || 100
 
-    {s1, s2} = time "Creating 2 #{size} element sets", fn ->
-      s1 = Enum.reduce(1..size, State.new(:s1), fn i, acc ->
+    {r1, r2} = time "Creating 2 #{size} element sets", fn ->
```
josevalim: I am almost sure `s1` and `s2` means state. :)

josevalim:
Thank you @folz! I agree this is an improvement. We shall only merge the first commit though. The short variables mean "module" and "state" respectively. :)

folz:
Thanks for your feedback and reviewing this PR!

chrismccord:
Please squash into a single commit

folz:
Done.

chrismccord:
I tweaked "replica" usage with "node" since we already use the term elsewhere with our "Replica" module that handles crdt replication. so primary/secondary works to describe erlang nodes and replicas remain isolated to crdt replication. Thanks!

PR #31, opened 2016-04-07
Optimize lookups with ets pids table

chrismccord:
Add pids table for reverse lookup against
values table for efficient matching of
get_by_pid.

Optimize ets select for get_by_topic
by or'ing downed node comparisons.

### DIFF
--- a/lib/phoenix_pubsub/tracker/state.ex
+++ b/lib/phoenix_pubsub/tracker/state.ex
@@ -4,24 +4,26 @@ defmodule Phoenix.Tracker.State do
   """
   alias Phoenix.Tracker.{State, Clock}
 
-  @type name      :: term
-  @type topic     :: String.t
-  @type key       :: term
-  @type meta      :: Map.t
-  @type ets_id    :: pos_integer
-  @type clock     :: pos_integer
-  @type tag       :: {name, clock}
-  @type cloud     :: MapSet.t
-  @type context   :: %{name => clock}
-  @type values    :: ets_id | :extracted | %{tag => {pid, topic, key, meta}}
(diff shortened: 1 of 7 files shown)


lib/phoenix_pubsub/tracker/state.ex
```diff
@@ -51,7 +54,8 @@ defmodule Phoenix.Tracker.State do
     reset_delta(%State{
       replica: replica,
       mode: :normal,
-      values: :ets.new(:values, [:duplicate_bag]),
+      values: :ets.new(:values, [:ordered_set]),
+      pids: :ets.new(:pids, [:duplicate_bag]),
```
indrekj: @chrismccord I know this is an old commit, but do you remember why did you use duplicate bag here instead of just bag which still allows duplicate keys?

chrismccord: duplicate bag will not check for existing keys on insert. It gave us a 10x speedup when we profiled phoenix pubsub back in 2015 for the 2M connection benchmark

chrismccord: Phoenix channels*. More info on the original PR:
https://github.com/phoenixframework/phoenix/pull/1311

Insertion time is constant for dup bag, where the insertion grows linearly with the bag size for an ordered set. I just realized you are asking `:bag` vs `:duplicate_bag`, not vs `:ordered_set`. I'm not sure of the differences, if any, for bag vs dup bag inserts in the above context.

indrekj: Ok. Thanks

PR #34, opened 2016-04-16
Add ability to intercept pubsub messages

chrismccord:
@josevalim this is an experiment from ideas we discussed earlier today. I realized we could support efficient message intercepting without a central process by making a simple callback module. See the `Phoenix.PubSub.Interceptor` moduledoc for example usage. Thoughts?

### DIFF
--- a/lib/phoenix_pubsub/interceptor.ex
+++ b/lib/phoenix_pubsub/interceptor.ex
@@ -1,0 +1,47 @@
+defmodule Phoenix.PubSub.Interceptor do
+  @moduledoc ~S"""
+  A behaviour for intercepting PubSub messages before they are relayed locally.
+
+  Allows broadcasted messages to be intercepted before being forwarded to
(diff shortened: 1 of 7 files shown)


josevalim:
I like this but I would shelve it for now (i.e. keep the PR open but not merge it) until we have more evidence we need it.

Also it is worth considering if we should have multiple interceptors, if we should call them interceptors, and if so, how we describe those interceptors compared to channel interceptors. For example, maybe the callback should also be called `handle_out/3`? I typically like to write documentation for those cases because if we cannot properly explain which one is which, we likely have the wrong name.

This PR also made me realize that we moved the fastlane configuration to the adapter: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/adapter.ex#L45

This will make it a bit annoying if we want to move the pubsub process out of the endpoint supervisor. Do you remember why did we decide that and are there other ways we can solve this?

PR #35, opened 2016-04-16
Set default pool size to number of schedulers

Gazler:
Maybe there is a better way to do the test?


Gazler:
Apparently closing a pull request, force pushing then re-opening it is disabled. I'll open a new one.

PR #36, opened 2016-04-16
Set default pool size to number of schedulers

Gazler:
Maybe there is a better way to do the test than checking that the last shard subscriber count?


josevalim:
@Gazler maybe you can ask the number of children to the supervisor?

Gazler:
@josevalim Let me know if you prefer that and I'll squash it down.

chrismccord:
<3 <3 <3

PR #38, opened 2016-05-12
No need to encode the random bytes in Erlang's external term format

MSch:
I think that's a leftover from changing from `make_ref()` to `:crypto.strong_rand_bytes(8)` in 31a68a2c.


jeregrine:
Thanks! ❤️

Issue #50, opened 2016-07-09
Fastlanes via subscribe


merqlove:
Can't find way to make it works.

I.e.:
```elixir
Phoenix.PubSub.subscribe(:freeling, "freeling:#{id}", fastlane: {some_module_pid, SomeModule, ["some_id1"]})
```

Normal subscriber always called, but not fastlane destination.

Also tried to print subscribers, all right, it has fastlanes:
```
[{#PID<0.959.0>,
  {#PID<0.1006.0>, SomeModule,
   ["bdb59e6a-2a5d-11e6-be56-685b35999bd2"]}}]
[{#PID<0.959.0>,
  {#PID<0.1007.0>, SomeModule,
   ["bdb59e6a-2a5d-11e6-be56-685b35999bd2"]}}]
```

Seems that sending to local pid is forced.
[lib/phoenix/pubsub/local.ex#L115](https://github.com/phoenixframework/phoenix_pubsub/blob/master/lib/phoenix/pubsub/local.ex#L115)

```elixir
  defp do_broadcast(nil, pubsub_server, shard, from, topic, msg) do
    pubsub_server
    |> subscribers_with_fastlanes(topic, shard)
    |> Enum.each(fn
      {pid, _} when pid == from -> :noop
      {pid, _} -> send(pid, msg)
    end)
  end
```

merqlove:
I'm asking this because in my case i can't send fastlane from broadcaster, but seems i can provide it at subscribe state..

merqlove:
Example fastlane calback is:

```elixir
def fastlane(msg) do
  IO.inspect msg
  msg
end
```

chrismccord:
Please verify the versions you are using, as well as the pubsub adapter. If using Redis please see if you can recreate with pg2. Failing that, a minimal example which recreates the issue would be very helpful. Thanks!

merqlove:
Hello, Chris, thanks for fast reply.

```elixir
[{:phoenix_pubsub, "~> 1.0"},     
 {:phoenix_pubsub_redis, "~> 2.1.2"},]
```

In current project we have Mongo as primary source, so it can be a bit complicated :) 

Seems that this problem not depends on `Redis` or any other PubSub source at all, because they always call `Phoenix.PubSub.Local.broadcast/6 ` to deliver it for local subscribers.

merqlove:
For example you did it in Redis PubSub:

```elixir
[...]

PR #51, opened 2016-07-09
Fix fastlane delivery for subscribers

merqlove:
Fix for #50 
Please correct me in the process if i did some mistake.
Added 1 line. Tests passed.

```elixir
  defp do_broadcast(nil, pubsub_server, shard, from, topic, msg) do
    pubsub_server
    |> subscribers_with_fastlanes(topic, shard)
    |> Enum.each(fn
      {pid, _} when pid == from -> :noop
      {_, {fastlane_pid, module, data}} -> module.fastlane(fastlane_pid, msg, data) # added
      {pid, _} -> send(pid, msg)
    end)
  end
```

### DIFF
--- a/lib/phoenix/pubsub/local.ex
+++ b/lib/phoenix/pubsub/local.ex
@@ -112,6 +112,7 @@ defmodule Phoenix.PubSub.Local do
     |> subscribers_with_fastlanes(topic, shard)
     |> Enum.each(fn
       {pid, _} when pid == from -> :noop
+      {_, {fastlane_pid, module, data}} -> module.fastlane(fastlane_pid, msg, data)
       {pid, _} -> send(pid, msg)
     end)
   end


merqlove:
Also found working way, without this pull request
created fastlane module, and added it to init of PubSub.

Seems it works like this:

```elixir
# lib/module_name.ex
supervisor(Phoenix.PubSub.Redis, [..., fastlane: ModuleName.Fastlane]]),
```

```elixir
# lib/module_name/fastlane.ex
defmodule ModuleName.Fastlane do
  def fastlane(subscribers, from, msg) do
    Enum.each(subscribers, fn
      {pid, _} when pid == from -> :noop
      {_, {fastlane_pid, module, data}} -> module.fastlane(fastlane_pid, msg, data)
      {pid, _} -> send(pid, msg)
    end)
  end
end
```

merqlove:
It will not work in one case:

1) We have defined `fastlane` in the init, but function argument is nil(`Local.broadcast(nil, ...)`).
To fix it at my side i have predefined `ModuleName.Fastlane` module name in my different app (which is not Erlang at all). 
But i feel its wrong. 
If i change this module name in the future, remote app will be unresponsive.
It must be at Phoenix PubSub side by checking default option if fastlane argument is nil.

merqlove:
This words about `master` branch, not about current PR

merqlove:
Test demonstrating this error(added similar state opts fetch as in Redis for PG2):
https://github.com/merqlove/phoenix_pubsub/commit/cd6d4ce45facb7e7f47dec5d8cc7070cae4dbe19

merqlove:
Same for redis.
For me it works in app, but library tests failed.
https://github.com/merqlove/phoenix_pubsub_redis/commit/0d4cbc4d8ef49956689092cab033a19aceea5be3
Currently it is one and only way, because Local supervisor didn't receiving all options.

PR #54, opened 2016-09-06
Fix documentation function references

binaryseed:
Fixes a reference to the untrack function, and removes a duplicated sentence...


josevalim:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #67, opened 2017-01-06
Allow getting tracker data by key

adkron:
`Phoenix.Tracker.get_by_key` allows us to request a specific piece of
data instead of traversing the whole list. I thought I would return an
empty map if we do not find the key in the list because it seemed like a
sane default to me. When I ask for a map I would like to receive one.

Amos King @adkron <amos@binarynoggin.com>

### DIFF
--- a/lib/phoenix/tracker.ex
+++ b/lib/phoenix/tracker.ex
@@ -101,6 +101,7 @@ defmodule Phoenix.Tracker do
 
   @type presence :: {key :: String.t, meta :: Map.t}
   @type topic :: String.t
+  @type key :: term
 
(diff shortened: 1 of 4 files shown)


adkron:
I'm trying to address https://github.com/phoenixframework/phoenix/issues/1777. I also think that this will allow the tracker to be used for any fault tolerant ephemeral data anyone would one to track outside of presence.

chrismccord:
Closed via https://github.com/phoenixframework/phoenix_pubsub/commit/267910395bac782d45754fe82f96cb2dfd8500b0

PR #72, opened 2017-04-11
Fixed type for Tracker.init callback

lasseebert:
The tracker implementation is supposed to return `{:ok, state}` and not `{:ok, pid}`.

This change fixes dialyzer warnings in my own project that uses phoenix_pubsub.


chrismccord:
❤️❤️❤️

Issue #76, opened 2017-05-09
Is a `:one_for_one` restart strategy safe for my supervisor?


myronmarston:
We're using this library in a non-phoenix project and are defining our supervision tree ourselves.  Here's the children I'm defining under a supervisor:

``` elixir
    [
      supervisor(Task.Supervisor, [[name: MyApp.PubSubTracker.TaskSupervisor]]),
      supervisor(Phoenix.PubSub.PG2, [MyApp.PubSub, []]),
      worker(MyApp.PubSubTracker, [[
        pubsub_server: MyApp.PubSub,
        name: MyApp.PubSubTracker,
        task_sup: MyApp.PubSubTracker.TaskSupervisor,
      ]]),
    ]
```

We're trying to figure out what restart strategy to use.  When I'm dealing with GenServers I've written I can generally figure out if `:one_for_one` is safe or if I need to use `:rest_for_one`.  Here I' not sure, because I don't know the details of how the tracker and pubsub processes interact.  Is it safe for them to be restarted individually (allowing us to use `:one_for_one`)?  Or do we need to use `:rest_for_one`?

If possible, it would be nice for this to be in the docs.

Thanks!

josevalim:
It is probably fine to be one_for_one because the pubsub tracker is simply going to crash if it tries to use the pubsub while it is down. Also if the tracker is subscribed to pubsub, it should automatically go down in case it crashes.

myronmarston:
Great, thanks.  We'll do that.

PR #77, opened 2017-05-12
Improve docs

myronmarston:
We started using this library recently, but found the docs confusing and/or inaccurate in some places.  I've taken a stab at improving them.  Some parts I wasn't 100% sure on so I might need to revert some commits.  (You may also not want the `.gitignore` and `mix.exs` changes, but they helped me work on the docs locally and seem generally useful for other contributors).

### DIFF
--- a/lib/phoenix/pubsub.ex
+++ b/lib/phoenix/pubsub.ex
@@ -41,11 +41,6 @@ defmodule Phoenix.PubSub do
 
       supervisor(Phoenix.PubSub.Redis, [:my_pubsub, host: "192.168.100.1"])
 
-  Note that a tracker and a pubsub server can generally be restarted
-  independently, so if they exist under the same supervisor, you can
-  safely use a `:one_for_one` restart strategy (unless your custom
-  tracker logic or additional supervised processes dictate otherwise).
-
   The configuration above will start a Redis pubsub and
   register it with name `:my_pubsub`.
 
--- a/lib/phoenix/pubsub.ex
+++ b/lib/phoenix/pubsub.ex
@@ -119,7 +119,7 @@ defmodule Phoenix.PubSub do
   ## Duplicate Subscriptions
 
   Callers should only subscribe to a given topic a single time.
-  Duplicate subscriptions for a Pid/topic pair are allowed and
+  Duplicate subscriptions for a pid/topic pair are allowed and
   will cause duplicate events to be sent; however, when using
   `Phoenix.PubSub.unsubscribe/3`, all duplicate subscriptions
   will be dropped.
@@ -153,7 +153,7 @@ defmodule Phoenix.PubSub do
   end
   @spec subscribe(atom, pid, binary, Keyword.t) :: :ok | {:error, term}
   def subscribe(server, pid, topic, opts) do
-    IO.write :stderr, "[warning] Passing a Pid to Phoenix.PubSub.subscribe is deprecated. " <>
+    IO.write :stderr, "[warning] Passing a pid to Phoenix.PubSub.subscribe is deprecated. " <>
                       "Only the calling process may subscribe to topics"
     call(server, :subscribe, [pid, topic, opts])
   end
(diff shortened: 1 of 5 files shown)


lib/phoenix/tracker.ex
```diff
@@ -110,8 +117,9 @@ defmodule Phoenix.Tracker do
   @doc """
   Tracks a presence.
 
-    * `server_name` - The registered name of the tracker server
-    * `pid` - The Pid to track
+    * `server` - The pid or registered name of the tracker server
+    * `pid` - The pid to track. The tracker will monitor the process and
+      and automatically untrack it when it exits.
```
myronmarston: I wasn't 100% sure on this, but decided to document my understanding of what `pid` is used for.

Actually, from the code, it looks like the processes are actually linked instead of monitored which is surprising.  The tracker server has called `Process.flag(:trap_exit, true)` so that exits or crashes from linked tracked processes are converted into messages instead of taking down the tracker (which is good) but I believe a crash of the tracker will take down a tracked process due to the bi-directional nature of the link.  That seems dangerous.  Do we really want tracked processes to be forced to exit if/when the tracker crashes?  Can the tracker server use monitoring instead of a link?

lib/phoenix/pubsub.ex
```diff
@@ -112,14 +117,14 @@ defmodule Phoenix.PubSub do
   @doc """
   Subscribes the caller to the PubSub adapter's topic.
 
-    * `server` - The Pid registered name of the server
+    * `server` - The registered name of the server
```
chrismccord: The Pid or registered name of the server

lib/phoenix/pubsub.ex
```diff
@@ -137,23 +142,23 @@ defmodule Phoenix.PubSub do
...
   def subscribe(server, topic, opts)
     when is_atom(server) and is_binary(topic) and is_list(opts) do
     call(server, :subscribe, [self(), topic, opts])
   end
+  @spec subscribe(atom, pid, binary) :: :ok | {:error, term}
+  def subscribe(server, pid, topic)
+    when is_atom(server) and is_pid(pid) and is_binary(topic) do
+    subscribe(server, pid, topic, [])
+  end
   @spec subscribe(atom, binary) :: :ok | {:error, term}
```
chrismccord: Why was this clause moved down?

lib/phoenix/tracker.ex
```diff
@@ -110,8 +117,9 @@ defmodule Phoenix.Tracker do
   @doc """
   Tracks a presence.
 
-    * `server_name` - The registered name of the tracker server
-    * `pid` - The Pid to track
+    * `server` - The pid or registered name of the tracker server
+    * `pid` - The pid to track. The tracker will monitor the process and
```
chrismccord: "The tracker will automatically cleanup the pid when it exits"

lib/phoenix/pubsub.ex
myronmarston: But pids aren't supported.  All of the non-deprecated function heads for `subscribe` have a `when is_atom(server)` guard clause.  Only atoms are supported.

lib/phoenix/pubsub.ex
myronmarston: Because of how it rendered in [the docs](https://hexdocs.pm/phoenix_pubsub/Phoenix.PubSub.html#subscribe/3):

screen shot 2017-05-16 at 7 15 33 am

The docs are rendering `subscribe(server, pid, topic)` as the main function head for `subscribe/3`.  It's quite confusing because then you spell out 3 arguments which are different (`server`, `topic`, and `opts`).  At first I was quite confusing, then thought there was a typo in the docs when I saw this.  After looking at the code I realized that there are two `subscribe/3` function heads.  However, the one that takes a `pid` as the 2nd argument is deprecated since it delegates to `subscribe/4` which prints a deprecation warning.  So that explains why `pid` isn't included in the list of described arguments.  But it also means that we don't want the deprecated function head to be the one that ExDoc renders.  So I moved it down so that it'll now render `subscribe(server, topic, opts)` as the main function head, which aligns with the args you describe.

Of course, moving the function clause ordering bears some risk of changing the semantics of the code, but the two function heads I re-ordered have guard clauses for there are arguments and are mutually exclusive, so no call could ever match both.  Therefore, the ordering isn't being used for semantics here.

myronmarston:
> I have a few small comments. Please also remove the note on the supervision strategies and we are good to go. Thanks!

Updated as requested.

I'm curious about the removal of the supervision strategies note--such a thing would have kept me from having to ask about it in #76. Was my note inaccurate or is it just not something you want the docs to explain?

chrismccord:
Sorry for the delay on this!  Can you take care of the conflicts and I will get this merged in. Thanks!

myronmarston:
Sorry, it's been so long since I wrote this that it's not clear to me how to resolve the merge conflicts. 

I'm just going to close this.

PR #78, opened 2017-06-08
Configurable broadcast strategy

pzel:
# Configurable Broadcast Strategies for `Phoenix.PubSub.PG2`

## Summary

This work is a result of load tests conducted on a Phoenix system with the 
following characteristics:

  * Many (100K—300K) channels connected via websockets

  * Short channel lifetimes (0—300sec)

  * Large, one-sided channel subscriptions (1 to 100—500)  
      Implemented as internal subscriptions from one channel to multiple topics,
      *not* as channel-per-subscription.

Under the given constraints, we observed that the `pix_lock` and `proc_link` 
VM locks were acquired very often, resulting in lock contention. The BEAM virtual
machine became increasingly sluggish and ultimately unrensponsive as the number
of broadcasts issued went up.

We have determined that the code responsible for grabbing `pix_lock` and `proc_link`
is called from the [broadcast/6 function in Phoenix.PubSub.Local][1]. At 
sufficiently high load, the cost incurred by `Task.async` overwhelms the gains
to be had from parallelized broadcast. Lock counter statistics are provided below.

This problem is addressed by the `Phoenix.PubSub.Strategy` behaviour, which allows
the developer to select the manner in which broadcasts will be sent out to all the PubSub
pool member shards. The `Parallel` strategy does exactly what the original implementation
did, while the `Serial` strategy performs the broadcast sequentially to the shards, in numeric
order.

Using the `Serial` strategy lets the Phoenix PubSub system handle much higher
message rates. At very high loads, using the `Parallel` strategy can bring down
the system via lock contention, while the `Serial` strategy leaves the system
responsive.

## Comments & Caveats

We believe that the `Serial` strategy makes more sense for most applications of
real-time messaging with Phoenix PubSub, but this PR maintains `Parallel` as the
default implementation. This preserves compatibility with the existing codebase.

Furthermore, only `Phoenix.PubSub.PG2` leverages the choice of strategy. The Redis
pubsub module remains unaffected by any configuration changes.

## BEAM VM lock statistics for Parallel and Serial strategies

The below measurements were taken on a BEAM VM (Erlang/OTP 19.3) compiled with lock
counting capabilities (`--enable-lock-counter`). The Phoenix system was running the
application described above, with 100K channels, ~1.3K broadcasts/sec, on a
36-vCPU machine, with a PubSub `pool_size` of 36.

#### `Phoenix.PubSub.Strategy.Parallel` (the original PubSub.Local default)

```
lock             id    #tries  #collisions  collisions [%]  time [us]  duration [%]
-----           ---   ------- ------------ --------------- ---------- -------------
run_queue        36 106927778      4038734          3.7771   16127265       53.7447
pix_lock       1024    178959         3318          1.8541    6858644       22.8567
proc_link    149494  21039238       420852          2.0003    6051021       20.1652
proc_msgq    149494  40909368       129044          0.3154    2312437        7.7063
pollset           1    939200        60844          6.4783     498849        1.6624
drv_ev_state     16   1617956        23982          1.4822     281005        0.9365
```

#### `Phoenix.PubSub.Strategy.Serial`

```
lock             id    #tries  #collisions  collisions [%]  time [us]  duration [%]
-----           ---   ------- ------------ --------------- ---------- -------------
run_queue        36 115589890      6773095          5.8596   33032272      110.1032
proc_link    148354  20171968       245398          1.2165    1103565        3.6784
pollset           1    848509        66201          7.8020     531351        1.7711
drv_ev_state     16   1176528        28127          2.3907     364193        1.2139
proc_main    148354  49105560       305810          0.6228     198287        0.6609
db_hash_slot   4800  21299225         8677          0.0407      83465        0.2782
         
```

As evidenced by the graphs, omitting parallelism during broadcast frees up the BEAM 
VM to run its regularly-scheduled processes (the `run_queue` lock).
This directly translates to application responsiveness and capacity to handle load.

#### `Parallel` strategy
image

#### `Serial` strategy
image

The graphs compare the same scenario executed using the `Parallel` and `Serial` strategies.   
The broadcast message rate was selected to show system behaviour under a heavy load,          
but not so heavy as to break the application.                                                 
                                                                                              
The system using the `Parallel` strategy is not completely unresponsive, but                  
some symptoms of overload are noticeable, such as higher tail latencies for                   
operations like Presence.track.                                                               
                                                                                              
The system using the `Serial` strategy displays much better tail latencies under              
the exact same load scenario.                                                                 

[1]: https://github.com/phoenixframework/phoenix_pubsub/blob/master/lib/phoenix/pubsub/local.ex#L99-L106

### DIFF
--- a/lib/phoenix/pubsub/local.ex
+++ b/lib/phoenix/pubsub/local.ex
@@ -92,18 +92,19 @@ defmodule Phoenix.PubSub.Local do
       :ok
 
   """
-  def broadcast(fastlane, pubsub_server, 1 = _pool_size, from, topic, msg) when is_atom(pubsub_server) do
+
+  def broadcast(fastlane, pubsub_server, pool_size, from, topic, msg,
+    strategy \\ Phoenix.PubSub.Strategy.Parallel)
+  def broadcast(fastlane, pubsub_server, 1 = _pool_size, from, topic, msg, _)
+    when is_atom(pubsub_server) do
     do_broadcast(fastlane, pubsub_server, _shard = 0, from, topic, msg)
     :ok
   end
-  def broadcast(fastlane, pubsub_server, pool_size, from, topic, msg) when is_atom(pubsub_server) do
-    parent = self()
-    for shard <- 0..(pool_size - 1) do
-      Task.async(fn ->
-        do_broadcast(fastlane, pubsub_server, shard, from, topic, msg)
-        Process.unlink(parent)
-      end)
-    end |> Enum.map(&Task.await(&1, :infinity))
+  def broadcast(fastlane, pubsub_server, pool_size, from, topic, msg,
+    strategy) when is_atom(pubsub_server) do
+    strategy.broadcast(pool_size, fn(shard) ->
+      do_broadcast(fastlane, pubsub_server, shard, from, topic, msg)
+    end)
     :ok
   end
 
--- a/lib/phoenix/pubsub/local.ex
+++ b/lib/phoenix/pubsub/local.ex
@@ -102,7 +102,7 @@ defmodule Phoenix.PubSub.Local do
   end
   def broadcast(fastlane, pubsub_server, pool_size, from, topic, msg,
     strategy) when is_atom(pubsub_server) do
-    strategy.broadcast(pool_size, fn(shard) ->
+    strategy.run(pool_size, fn(shard) ->
       do_broadcast(fastlane, pubsub_server, shard, from, topic, msg)
     end)
     :ok
--- a/lib/phoenix/pubsub/pg2.ex
+++ b/lib/phoenix/pubsub/pg2.ex
@@ -33,9 +33,13 @@ defmodule Phoenix.PubSub.PG2 do
   def init([server, opts]) do
     scheduler_count = :erlang.system_info(:schedulers)
     pool_size = Keyword.get(opts, :pool_size, scheduler_count)
+    broadcast_strategy = Keyword.get(opts, :broadcast_strategy,
+      Phoenix.PubSub.Strategy.Parallel)
     node_name = opts[:node_name]
-    dispatch_rules = [{:broadcast, Phoenix.PubSub.PG2Server, [opts[:fastlane], server, pool_size]},
-                      {:direct_broadcast, Phoenix.PubSub.PG2Server, [opts[:fastlane], server, pool_size]},
+    dispatch_rules = [{:broadcast, Phoenix.PubSub.PG2Server,
+                       [opts[:fastlane], server, pool_size, broadcast_strategy]},
+                      {:direct_broadcast, Phoenix.PubSub.PG2Server,
+                       [opts[:fastlane], server, pool_size, broadcast_strategy]},
                       {:node_name, __MODULE__, [node_name]}]
 
     children = [
--- a/lib/phoenix/pubsub/pg2.ex
+++ b/lib/phoenix/pubsub/pg2.ex
@@ -22,6 +22,11 @@ defmodule Phoenix.PubSub.PG2 do
       topic or greater than 1M clients, a pool size equal to the number of
       schedulers (cores) is a well rounded size.
 
+    * `:broadcast_strategy` - What strategy to use for broadcasting messages.
+      Currently, `Phoenix.PubSub.Strategy.Parallel` is the default implementation.
+      For systems with very high broadcast throughput,
+      `Phoenix.PubSub.Strategy.Serial` may yield better performance.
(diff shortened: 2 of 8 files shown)


josevalim:
Thank you @pzel! I would like to propose another contract, one that basically replaces `Task` by a custom module. And your custom module could be this:

```elixir
defmodule InlineTask do
  def async(fun) do
    fun.()
  end

  def await(result, _timeout) do
    result
  end
end
```

The reasoning is that it is less dependent on the internals and less relying on shards and pool sizes.

josevalim:
Actually, hold that thought. The Task one is not going to work if we are doing operations like Process.unlink.

josevalim:
Ok, here is another approach. What if we allow customizing the number of tasks used for dispatching? This way we don't need any strategy or callbacks, we just keep it in as a regular option.  It can be called `:max_dispatchers` or `:max_broadcasters`. Thoughts?

josevalim:
Ok, passing max_dispatchers will be annoying to implement because, what if you specify 5 broadcasters and you have 8 partitions? Back to the drawing board, haha.

josevalim:
Ok, the task approach is going to work since we can call `Process.unlink(self())`. After this back and forth, that would be my pick.

Whatever we decide here we need to replicate in Elixir's Registry. As we have a new PubSub implementation based on top of Elixir's Registry.

pzel:
Hi @josevalim, thanks for the interest and encourgement! I apologize for the
delayed response -- I'm away on vacation all week. Just so that I understand
correctly what you're proposing, let me recap my understanding of it:

We'd want to have the implementation of `Phoenix.Pubsub.Local.broadcast/7,8`
follow something along the lines of:

```
task_impl = @configured_task_impl # passed in as parameter, macro, whatever
parent = self()
for shard <- 0..(pool_size - 1) do
  task_impl.async(fn ->
    do_broadcast(fastlane, pubsub_server, shard, from, topic, msg)
    Process.unlink(parent)
  end)
  end |> Enum.map(&task_impl.await(&1, :infinity))
:ok
```

where ```task_impl = Task | InlineTask ```as defined by you in the comment above. 

That way, we can ditch the whole `Strategy.*` business and avoid a lot of extra
code in the codebase.

In general, I like the idea very much and I'm open to amending the code to use
this pattern. There are 2 things that I find a bit inelegant:

  1) Having to call `unlink(self())` for every `do_broadcast`. In our use case,
  this will be million(s) of times per second, potentially.
  
  2) The `Task`-able(?) interface being named `async/1` and
  `await/2`. Specifically, whether the `Task` at hand is `async` or not becomes
  an implementation detail.  Perhaps this is a good point in time to reconsider
  the behaviour and change the function names to `Task.launch/1` and
  `Task.collect/2`, or something similar?
  
## The counter-idea

On the other hand, the brutalist in me says that we should just drop the whole
abstraction and **always** just loop over the shards. Message sending time in
the VM is negligible, and certainly lower than the overhead of spawning
processes. This prevents users from hitting performance problems when their
userbase/traffic grow.

This gives us all the right benefits:

    1) less code in the codebase
    2) no need for an esoteric config option
    3) performant in all scenarios especially the high-load ones

What would it take to convince you to accept this approach? ;) This entire PR in
that case would consist of removed lines, with just this bit changed:

```
for shard <- 0..(pool_size - 1) do
    do_broadcast(fastlane, pubsub_server, shard, from, topic, msg)
  end)
:ok
```

josevalim:
We can ditch the parallelism but we need to consider other workflows. The code today is based on the assumption that the parallelism is beneficial when dispatching to multiple users. How many users do you have on each channel?

pzel:
Our use case is 0 - 500 subscriptions per topic. That's probably why we've yet to encounter the kind of load at which parallelism pays its bills.

I think you're right: without benchmarks showing the 'tipping point', we can't make a strong case for either strategy being the only available implementation.

josevalim:
To be fair, I would expect that to be the most common workload. Most applications will be exercising parallelism through multiple publishers. I would say the approach we have today will only be better when you have a single publisher publishing large amounts of data.

chrismccord:
Thanks for this PR and for taking the time to instrument the implementations! I agree with your brutalist side that the simplest approach forward that is performant on all usecases is to simply make things serial. We can reconsider parallelized broadcasts when or if we have folks hit perf issues in the other direction. Can you update the PR? Thanks!

pzel:
Hi @chrismccord, please see #82 for the much-simplified version. :)

chrismccord:
Thanks so much! Merged. We can revisit these changes if strategies become needed in the future. Thanks!

PR #80, opened 2017-06-11
Fix presence ghosting

chrismccord:
@gazler this includes the test cases from your branch that re-create the ghosting issues. I've changed the way replication is done to only extract and transfer presences within the remote node's context, so we avoid replicating presences that may be raced in a permdown. Everything looks good on my end with my own applications. Can you run it through its paces locally and verify on your end with the scenarios you've done preivously? Thanks!

Chris

### DIFF
--- a/lib/phoenix/tracker/delta_generation.ex
+++ b/lib/phoenix/tracker/delta_generation.ex
@@ -8,15 +8,15 @@ defmodule Phoenix.Tracker.DeltaGeneration do
 
   Falls back to extracting entire crdt if unable to match delta.
   """
-  @spec extract(State.delta, [State.delta], State.context) :: State.delta | State.t
-  def extract(%State{mode: :normal} = state, generations, remote_context) do
+  @spec extract(State.delta, [State.delta], State.name, State.context) :: State.delta | State.t
+  def extract(%State{mode: :normal} = state, generations, remote_ref, remote_context) do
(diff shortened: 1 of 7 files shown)


Gazler:
@chrismccord I'll check this out tomorrow.

lib/phoenix/tracker/delta_generation.ex
```diff
@@ -54,11 +54,16 @@ defmodule Phoenix.Tracker.DeltaGeneration do
...
   end
 
-  defp delta_fullfilling_clock(generations, remote_clock) do
+  defp delta_fullfilling_clock(generations, remote_context) do
     generations
     |> Enum.with_index()
-    |> Enum.find(fn {%State{range: {local_start, _local_end}}, _} ->
-      Clock.dominates?(remote_clock, local_start)
+    |> Enum.find(fn {%State{range: {local_start, local_end}}, _} ->
+      if Clock.dominates_or_equal?(remote_context, local_start) and
```
Gazler: No need for the if here. Could be rewritten as:

Clock.dominates_or_equal?(remote_context, local_start) and
   not Clock.dominates?(remote_context, local_end)

chrismccord: Indeed. I was doing some inspection here but can be killed now

PR #83, opened 2017-07-14
Fix Dialyzer warnings

take-five:
After running Dialyzer on Elixir 1.4/OTP 20 with `--unknown` flag,
39 warnings have been emitted. This commit fixes all those warnings,
although it required introducing private type `Tracker.state`.

### DIFF
--- a/lib/phoenix/tracker/replica.ex
+++ b/lib/phoenix/tracker/replica.ex
@@ -3,13 +3,13 @@ defmodule Phoenix.Tracker.Replica do
   alias Phoenix.Tracker.Replica
 
   @type name :: String.t
-  @type vsn :: term
+  @type vsn :: integer
   @type replica_ref :: {name, vsn}
 
   @type t :: %Replica{
     name: name,
     vsn: vsn,
-    last_heartbeat_at: pos_integer,
+    last_heartbeat_at: pos_integer | nil,
     status: :up | :down | :permdown
   }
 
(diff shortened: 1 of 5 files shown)


lib/phoenix/tracker/replica.ex
```diff
@@ -77,9 +77,13 @@ defmodule Phoenix.Tracker.Replica do
     %Replica{replica | last_heartbeat_at: now_ms()}
   end
 
-  defp now_ms, do: System.system_time(:milli_seconds)
+  @otp_vsn :erlang.system_info(:otp_release) |> :erlang.list_to_integer()
+  @time_unit_ms if @otp_vsn < 19, do: :milli_seconds, else: :milliseconds
+  @time_unis_us if @otp_vsn < 19, do: :micro_seconds, else: :microseconds
```
chrismccord: Is this necessary? I believe Elixir should take care of this within `System`, no?

chrismccord:
Thanks! I have just one comment

lib/phoenix/tracker/replica.ex
take-five: > Is this necessary? I believe Elixir should take care of this within System, no?

Yes, it takes care of it and translates `:milli_seconds` to `:milliseconds`, but OTP 18 has different typespecs for time units than OTP 19 (http://devdocs.io/erlang~18/erts-7.3/doc/html/erlang#type-time_unit), hence this small hack

chrismccord:
❤️❤️❤️🐥🔥

PR #87, opened 2017-08-21
Update subscription documentation to not require pid

tsubery:
Avoid a deprecation warning when copy pasting from the documentation


chrismccord:
❤️❤️❤️🔥🐥

PR #88, opened 2017-09-01
fix `phoenix.pub_sub.bench.exs`

Ljzn:
Hi.

I found there is some error when I tried to run `mix phoenix.pub_sub.bench`. 

```
** (UndefinedFunctionError) function Phoenix.Tracker.State.extract/1 is undefined or private. Did you mean one of:

      * extract/3
```

After changing some code, it works. But I'm not sure is it correct.


chrismccord:
❤️❤️❤️🐥🔥

PR #95, opened 2018-01-17
start adding child_spec to adapters

pragdave:
Chris:

This is just a pathfinder to see if you're happy. If  so, I'll do it for redis, too.

Dave

### DIFF
--- a/lib/phoenix/pubsub/pg2.ex
+++ b/lib/phoenix/pubsub/pg2.ex
@@ -10,6 +10,26 @@ defmodule Phoenix.PubSub.PG2 do
         pubsub: [name: MyApp.PubSub,
                  adapter: Phoenix.PubSub.PG2]
 
+  To use this adapter outside of Phoenix, you must start an instance of 
+  this module as part of your supervision:
+
+      children = [
+        { Phoenix.PubSub.PG2, { name, options...} },
+
+        # or
+
+        { Phoenix.PubSub.PG2, name },
+
+        ...
+      ]
+
+  For example
+
+      children = [
+        { Phoenix.PubSub.PG2, { :connector, pool_size: 4 }},
+      ]
+
+
   ## Options
 
     * `:name` - The registered name and optional node name for the PubSub
@@ -23,6 +43,19 @@ defmodule Phoenix.PubSub.PG2 do
       schedulers (cores) is a well rounded size.
 
   """
+
+  def child_spec({name, options}) when is_list(options) do
+    %{
+      id:     __MODULE__,
+      start: { __MODULE__, :start_link, [ name, options] },
+      type:  :supervisor
+    }
+  end
+
+  def child_spec(name) do
+    child_spec({name, []})
+  end
+  
 
   def start_link(name, opts) do
     supervisor_name = Module.concat(name, Supervisor)
--- a/test/phoenix/pubsub/local_test.exs
+++ b/test/phoenix/pubsub/local_test.exs
@@ -22,6 +22,29 @@ defmodule Phoenix.PubSub.LocalTest do
             pool_size: size}}
   end
 
+  describe "PG2 child_spec" do
+    alias Phoenix.PubSub.PG2
+  
+    test "with just a name returns the name" do
+      spec = PG2.child_spec(:fred)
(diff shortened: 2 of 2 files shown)


chrismccord:
Thanks! Please go ahead with redis! ❤️❤️❤️❤️🐥🔥

josevalim:
@pragdave @chrismccord what do you think about only passing keywords, this way we don't need to nest tuples?

    {Phoenix.PubSub.PG2, name: Foo, ...}

We just need to pop the name and raise if it doesn't exist.

Thoughts?

pragdave:
I felt the API should look the same as start_link.

josevalim:
@pragdave yeah, that's a good point.

Although @pragdave, I have to point out you would probably be the first person to tell me we should let go of the "past" and go with what we think looks best, regardless of how it was before. :) I am not saying the keyword approach is best, I am just saying we probably don't need to get attached to the previous API.

For the Registry in Elixir, we went with the `{Module, opts}` route and even added a `start_link/1` to mirror it. YMMV.

chrismccord:
I'm fine with the keyword route as well, but I think we should add a /1 clause for start_link if we go that route since converting from the old style to the new child spec will be least surprising if the options match. My only hesitation is I went start_link/2 because `name` is always required so making it part of the keyword option lessens its significance a bit. Of course, Keyword.fetch! solves it so I think it would be fine.

pragdave:
Lol.

One of my concerns is that there's no obvious place to look for the
documentation of how to start something using the new supervisors. So
copying start_link seemed reasonable as a convention.

Although I'm the first person to agree that a nested tuple looks ugly,
there's another side to the coin. Keyword lists are typically used for
options; they stop you having to have functions with 17 arguments when you
only want to specify two of them. But there name is not an option. It's
required. So there's a strong indication that it should be it's own
parameter.


On Jan 17, 2018 09:08, "José Valim" <notifications@github.com> wrote:

@pragdave <https://github.com/pragdave> yeah, that's a good point.

Although @pragdave <https://github.com/pragdave>, I have to point out you
would probably be the first person to tell me we should let go of the
"past" and go with what we think looks best, regardless of how it was
before. :) I am not saying the keyword approach is best, I am just saying
we probably don't need to get attached to the previous API.

For the Registry in Elixir, we went with the {Module, opts} route and even
added a start_link/1 to mirror it. YMMV.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/phoenixframework/phoenix_pubsub/pull/95#issuecomment-358333422>,
or mute the thread

josevalim:
Keywords can also be used to give "names" to arguments, especially when it may be unclear what they are without the option. For example:

    {Phoenix.PubSub.PG2, MyApp.Foo}

vs:

    {Phoenix.PubSub.PG2, name: MyApp.Foo}

pragdave:
True. But then we'd have


{ child: Phoenix.PubSub.PG2, name: :Fred }


At which point someone would ask why user a tuple?


I actually originally wrote the PR using a name: option, then changed it.


I'm happy to change it back. Someone make the call.



On Wed, Jan 17, 2018 at 9:33am, José Valim < notifications@github.com [notifications@github.com] > wrote:
Keywords can also be used to give "names" to arguments, especially when it may be unclear what they are without the option. For example:

{Phoenix.PubSub.PG2, MyApp.Foo}


vs:

{Phoenix.PubSub.PG2, name: MyApp.Foo}

pragdave:
Is there a decision on this?

josevalim:
@pragdave agreed. :)

I will gladly delegate the decision to @chrismccord. :D

chrismccord:
Let's go with the keyword list and we'll make sure to raise if name isn't provided. I'll make the change here, meanwhile @pragdave can you send a PR based on this to phoenix_pubsub_redis? Thanks!

pragdave:
so start_link becomes just options too?

chrismccord:
yes, but for backwards compatibility we need to keep the 2-arity clause. Currently pubsub adapters are expected to be start_link/2

pragdave:
Yup: I was going to do that for both adapters.

PR #101, opened 2018-04-17
Optimise state

michalmuskala:
This pull request optimises the CRDT operations. 

The minor optimisations are based on some compiler tricks and using literal map pattern matching instead of functions in `Map`.

The bigger optimisations are based on smarter ets traversals - in `extract` and `merge` instead of folding over the whole table, we fold over only the relevant replica - this is still a full table scan, but a large part of filtering happens inside ets in C and not in Elixir, which also avoids a lot of copies.

Preliminary benchmarks show `extract` to be 10x faster and `merge` to be 25% faster (using the built-in benchmark).

### DIFF
--- a/lib/phoenix/tracker/state.ex
+++ b/lib/phoenix/tracker/state.ex
@@ -199,20 +199,23 @@ defmodule Phoenix.Tracker.State do
   def extract(%State{mode: :normal, values: values, clouds: clouds} = state, remote_ref, remote_context) do
     pruned_clouds = Map.take(clouds, Map.keys(remote_context))
     # fn {{topic, pid, key}, meta, {replica, clock}} when replica !== remote_ref ->
-    #  {replica, {replica, clock}, {pid, topic, key, meta}}
+    #  {{replica, clock}, {pid, topic, key, meta}}
     # end
-    ms = [
-      {{{:"$1", :"$2", :"$3"}, :"$4", {:"$5", :"$6"}},
-       [{:"=/=", :"$5", {:const, remote_ref}}],
-       [{{:"$5", {{:"$5", :"$6"}}, {{:"$2", :"$1", :"$3", :"$4"}}}}]}
-    ]
-    map =
-      for {replica, tag, data} <- :ets.select(values, ms),
-        Map.has_key?(remote_context, replica),
-        into: %{},
-        do: {tag, data}
+    ms = [{
+      {{:"$1", :"$2", :"$3"}, :"$4", {:"$5", :"$6"}},
+      [{:"=/=", :"$5", {:const, remote_ref}}],
+      [{{{{:"$5", :"$6"}}, {{:"$2", :"$1", :"$3", :"$4"}}}}]
+    }]
+    data =
+      foldl(values, [], ms, fn {{replica, _} = tag, data}, acc ->
+        if match?(%{^replica => _}, remote_context) do
+          [{tag, data} | acc]
+        else
+          acc
+        end
+      end)
 
-    {%State{state | clouds: pruned_clouds, pids: nil, values: nil, delta: :unset}, map}
+    {%State{state | clouds: pruned_clouds, pids: nil, values: nil, delta: :unset}, Map.new(data)}
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


lib/phoenix/tracker/state.ex
```diff
@@ -198,15 +198,19 @@ defmodule Phoenix.Tracker.State do
...
-          [{tag, {pid, topic, key, meta}} | acc]
-        else
-          acc
-        end
-    end) |> :maps.from_list()
+    # fn {{topic, pid, key}, meta, {replica, clock}} when replica !== remote_ref ->
+    #  {replica, {replica, clock}, {pid, topic, key, meta}}
+    # end
+    ms = [
+      {{{:"$1", :"$2", :"$3"}, :"$4", {:"$5", :"$6"}},
```
jeregrine: Wow this syntax is pretty wild. TIL about :ets.select matchspecifications.

lib/phoenix/tracker/state.ex
```diff
@@ -346,19 +365,15 @@ defmodule Phoenix.Tracker.State do
...
-        1 = :ets.select_delete(pids, [{{pid, topic, key}, [], [true]}])
-        delete_tag(acc, tag)
-      end)
-      |> Map.delete(replica)
+    # fn {{topic, pid, key}, _, {^replica, _}} -> {{pid, topic, key}, [], [true]} end
+    ms = [{{{:"$1", :"$2", :"$3"}, :_, {replica, :_}}, [], [{{{{:"$2", :"$1", :"$3"}}, [], [true]}}]}]
 
+    delete_ms = :ets.select(values, ms)
+    expected_deletes = length(delete_ms)
+    ^expected_deletes = :ets.select_delete(pids, delete_ms)
```
josevalim: Why do we do this match here? Are we trying to avoid a race?

lib/phoenix/tracker/state.ex
```diff
@@ -252,14 +266,21 @@ defmodule Phoenix.Tracker.State do
...
     unioned_clouds = union_clouds(local, remote)
-    init = {unioned_clouds, local.delta, []}
-
-    foldl(local.values, init, fn {{topic, pid, key}, _, tag} = el, {clouds, delta, leaves} ->
-      if in?(remote, tag) and not Map.has_key?(remote_map, tag) do
-        1 = :ets.select_delete(values, [{el, [], [true]}])
-        1 = :ets.select_delete(pids, [{{pid, topic, key}, [], [true]}])
+    %State{context: remote_context, clouds: remote_clouds, replica: replica} = remote
+    init = {unioned_clouds, delta, []}
+    # fn {key, meta, {^replica, clock}} -> {key, meta, {replica, clock}} end
```
josevalim: In the previous code we didn't match on the replica. Are you matching it here because `in?` is going to filter on it anyway?

lib/phoenix/tracker/state.ex
michalmuskala: I just replicated the previous behaviour that would assert on each `:ets.select_delete` call - it's very likely we don't need that at all.

lib/phoenix/tracker/state.ex
michalmuskala: That's what provides the improvement and leverages the fact that our CRDT is special in that we'll only get downs from that particular replica when processing the delta from that replica.

josevalim: Beautiful!

lib/phoenix/tracker/state.ex
josevalim: I would nuke it then. :)

lib/phoenix/tracker/state.ex
```diff
@@ -196,17 +198,24 @@ defmodule Phoenix.Tracker.State do
...
+      [{{{{:"$5", :"$6"}}, {{:"$2", :"$1", :"$3", :"$4"}}}}]
+    }]
+    data =
+      foldl(values, [], ms, fn {{replica, _} = tag, data}, acc ->
+        if match?(%{^replica => _}, remote_context) do
+          [{tag, data} | acc]
         else
           acc
         end
-    end) |> :maps.from_list()
```
michalmuskala: @josevalim @chrismccord there is something weird going on in that function even before extracting.

When called in tests `remote_ref` is an atom, but the `replica` value is a pair `{atom, integer}` so it can never match - keys in the `remote_conetxt` map have a similar format.

This makes me wonder what is broken - the implementation that assumes `replica` and `remote_ref` might match or tests.

michalmuskala: I pushed a commit that makes correct use of `State.extract` in tests. I'm not sure why the tests did not fail previously and if we're missing some test to check this particular behaviour.

chrismccord:
❤️❤️❤️🐥🔥

Issue #94, opened 2018-01-17
Should the adapters implement child_spec?


pragdave:
as Supervisor.Spec is deprecated?

chrismccord:
Yes indeed! Would you like to send a PR?

pragdave:
I saw what you did just then....  :)

PR #102, opened 2018-04-18
replace Supervisor.Spec with child_spec for PG2

pragdave:
Addresses issue #94 

Not sure how to test it directly, but it seems to be indirectly tested thanks to the change in Phoenix.Supervise.

Dave

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,20 +16,23 @@ def application do
   [applications: [:phoenix_pubsub]]
 end
 ```
-        
+
 ## Initialization (without Phoenix)
 ```elixir
 defmodule MyApp do
   use Application
 
   def start(_type, _args) do
-    import Supervisor.Spec, warn: false
 
     children = [
-      supervisor(Phoenix.PubSub.PG2, [MyApp.PubSub, []])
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -16,20 +16,23 @@ def application do
...
 ```elixir
 defmodule MyApp do
   use Application
 
   def start(_type, _args) do
-    import Supervisor.Spec, warn: false
 
     children = [
-      supervisor(Phoenix.PubSub.PG2, [MyApp.PubSub, []])
+      { Phoenix.PubSub.PG2, [ MyApp.PubSub, [] ]}
```
michalmuskala: I believe the preferred style would be without spaces around `{}[]`

michalmuskala:
The project nominatively still supports elixir 1.2, so it shouldn't use the new child-specs internally without bouncing that requirement.

README.md
pragdave: Might well be, but I don't use the formatter on religious grounds :)

josevalim: This project doesn't use the formatter. Even though, removing the spaces would make it consistent with the remaining code in the repository.

PR #111, opened 2018-08-28
Improve replication over netsplit

studzien:
This is done on top of #108 and #109.

This PR is an attempt to fix issues discovered by the property based tests that can generate split and join network between certain nodes:

1. [This counterexample](https://github.com/distributed-owls/breaking-pp/blob/master/test/large/cluster_counterexamples_test.exs#L66-L95) reproduced a case when joins are propagated via netsplit (node 1 stores information about sessions connected to node 3 internally), but leaves are not -- when the node 3 sessions are disconnected and the split is healed, these sessions still appear to be connected on node 1.
The proposed solution makes the filter in `observe_removes` less strict. I'm not sure what performance implications this might have, though.

2. [This counterexample](https://github.com/distributed-owls/breaking-pp/blob/master/test/large/cluster_counterexamples_test.exs#L97-L153) occasionally reproduces a situation when some of the sessions connected to node 3 never appear on node 2.
It seems that sometimes a merge of two deltas was possible, when their ranges were not contiguous. The proposed solution makes the deltas adjacency check more strict.

cc @pzel

### DIFF
--- a/lib/phoenix/tracker/clock.ex
+++ b/lib/phoenix/tracker/clock.ex
@@ -71,6 +71,15 @@ defmodule Phoenix.Tracker.Clock do
     Map.merge(c1, c2, fn _, v1, v2 -> min(v1, v2) end)
   end
 
+  @doc """
+  Returns the clock with just provided replicas.
+  """
+  def filter_replicas(c, replicas), do: Map.take(c, replicas)
+
+  @doc """
+  Returns replicas from the given clock.
+  """
+  def replicas(c), do: Map.keys(c)
 
   defp filter_clocks(clockset, {node, clock}) do
     clockset
--- a/lib/phoenix/tracker/delta_generation.ex
+++ b/lib/phoenix/tracker/delta_generation.ex
@@ -58,6 +58,9 @@ defmodule Phoenix.Tracker.DeltaGeneration do
     generations
     |> Enum.with_index()
     |> Enum.find(fn {%State{range: {local_start, local_end}}, _} ->
+      local_start = Clock.filter_replicas(local_start, Clock.replicas(remote_context))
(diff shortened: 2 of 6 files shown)


chrismccord:
Thank you so much! ❤️❤️❤️🐥🔥

studzien:
Our pleasure!
There will be one more coming up soon, fixing an initial sync race condition when more nodes are started, but there's still something failing the property tests sometimes in a non-deterministic way.

So most likely there will be something wrong with the test environment itself, but I want to make sure first before creating a PR :)

chrismccord:
Great! I have pushed 1.1.1 with your fixes and look forward to the next one! Thanks for patiently waiting for me to get these merged in. I have to devote the brain cycles anytime I revisit the tracker code :)

Thanks again!

studzien:
@chrismccord Actually the fix I've been referring to above has been already included in this PR :)

I've finally found the reason why the test harness was failing sometimes - simply the time we've been waiting for nodes to synchronize was too short given the Erlang Distributed failure detector time to detect that remote nodes were not reachable via netsplit.

It seems that current `master` passes the property tests we currently have.
I'm planning to improve their speed significantly, so it's feasible to run longer scenario that we currently do.

If you're interested, I've reorganized counterexamples in our repo and split them into ones caused by issues when [starting and stopping the nodes](https://github.com/distributed-owls/breaking-pp/blob/master/test/large/cluster_join_leave_counterexamples_test.exs) and the ones related to [netsplits](https://github.com/distributed-owls/breaking-pp/blob/master/test/large/cluster_netsplit_counterexamples_test.exs).

Each of them has a quick note what the issue was related to, and a link to a commit that fixed it.

PR #119, opened 2018-11-07
Fix documentation (DocTest) for Phoenix.PubSub

EskiMag:
Use `Process.info(self(), :messages)` instead of `Process.info(self())[:messages]`.
The later one was returning an empty array for me every time.

Issue appeared with Phoenix PubSub 1.1.1, Elixir 1.7.4, Erlang/OTP 21 [erts-10.1.1]


michalmuskala:
Ah, yes. That's caused by the change in OTP 21 where the `:messages` key is not returned by default from `Process.info/1` due to potentially large size.

PR #127, opened 2019-07-15
Optimize Shard.list and Shard.get_by_key

### LINKED ISSUE #124: Providing custom `Tracker.list` timeout
`Tracker.Shard` uses a `GenServer.call` which doesn't accept any options from the caller. This call can be expensive in a large shard or one which is under heavy load.

Is it worthwhile to expose a timeout option the entire way through the function calls? My thought is to set a low timeout value when writing a `listeners?` function and then just assuming true if the function times out.

I'm happy to add this in but wanted to run it by the maintainers first. I could see a desire to expose this through all of the Tracker public APIs.

indrekj:
Previous list and get_by_key had to go through GenServer to acquire
values ets table and replicas information. In case GenServer was
processing an update (e.g. heartbeat, track, untrack) then list and
get_by_key functions were blocked until it was completed. We saw this
behaviour in our cluster where simple list/get_by_key calls were
sometimes taking over few hundred milliseconds.

Storing replicas information in an ets table allows us to avoid going
through genserver and allows us to process list/get_by_key immediately.

I removed dirty_list function which was not public / exposed and which
was trying to resolve the same issue. dirty_list was called dirty
because it didn't check for down_replicas. This solution checks
down_replicas and doesn't change the api interface.

This should also resolve #124


indrekj:
Tried this out in production (patched v1.1.2). Forwarded ~15% traffic to the patched instance. Given endpoint does 1-30 `get_by_key` calls depending on the input. Here are the results:
Screen Shot 2019-08-03 at 11 33 59
First two graphs are from the v1.1.2 instances and the last one is from the patched one.

It looks faster, I'd say 20-50% faster. I was actually expecting better results but still better than before.

indrekj:
@josevalim what do you think?

indrekj:
@chrismccord any chance of getting this reviewed / merged as well?

luislhsc:
Any updates on this?

jaybe78:
I've been experiencing that issue where Phoenix.list timeout while all my channels topic are capped to 70 users max.

I've done a simple test with 50_000 joins and it times out at around 30K 

```
1..50_000 |> Enum.each(fn user ->
      user_pid =
      ChannelsHelpers.join_channel_in_task(%{
        test_pid: test_pid,
        user_id: "#{user}",
        channel: "challenges:rooms:lobby:#{channel_name}"
      })
    end)
```

It works if I get rid of this line in my channel 

```
push(socket, "presence_state", ChannelsPresence.list(room))
```

If the modifications committed improve the issue, is there any chance to merge those anytime soon ?

PR #130, opened 2019-09-04
Improve random_ref() function performance

indrekj:
This is called each time there's a join or other kind of update.

This is based on elixir plug [request id][1].

[1]: https://github.com/elixir-plug/plug/issues/629


chrismccord:
❤️❤️❤️🐥🔥

PR #133, opened 2019-10-30
Extend typespecs with no_return where function can raise

preciz:
Thanks for reviewing :wave: .


wojtekmach:
most functions will raise if given incorrect arguments and we're not annotating them as `no_return`; we reserve that type for functions that _never_ return (always raise, loop indefinitely, etc)

preciz:
Thank you for the clarification!

PR #134, opened 2019-11-11
Use random_ref instead of make_ref for transfer_req

spscream:
Hi, 

We have own implementation of direct_broadcast protocol and we have a bug in production with a transfer of transfer_req presence events. 

Server1 when sending does:
```
binary_transfer_req_msg = 
  transfer_req_msg
  |> :erlang.term_to_binary()
  |> Base.encode64!()
```
Server2 when receiving event:
```
event = 
  binary_transfer_req_msg
  |> Base.decode64!()
  |> :erlang.binary_to_term([:safe])
```
Now I'm getting ArgumentError since references can't be safely deserialized. The solution is to use something more safe for this ref. By the way, what is the purpose of this ref? It isn't used in transfer_ack.


spscream:
hm, this change also doesn't help in our case :( Closing request.

PR #136, opened 2019-11-28
Fix transfer req loop when a node is replaced

### LINKED ISSUE #135: Network traffic up for 20 minutes after a server restart
We're using Phoenix Presence (latest version in the master branch). We have a kubernetes set up where we have multiple pods running.

We've noticed that every time we restart a pod or do a rolling update then network traffic is up for 20 minutes.

I was able to replicate it in our beta environment when I had 10K online connections and I restarted one pod:
Screen Shot 2019-11-27 at 18 05 28
As you can see, traffic went up around 11:53 and came back down around 12:14.

I think it's related to `permdown_period` setting which by default is 20 minutes. I tried to replicate this with just phoenix_pubsub library without a web server but wasn't able to. EDIT: It is related. If I changed it to 10, then network traffic was up only for 10 minutes.

I also did a tcpdump inside one pod to see where the traffic is coming from/going to. It was all between the presence servers themselves. I think these are the state synchronization messages.

Do you have any suggestions what to look for or how to gather more information?

indrekj:
Scenario how Node2 is replaced by Node3 (this is also basically a
rolling update):
1. Node1 and Node2 are up and synced
2. Kill Node2 (node1 will start permdown grace period for Node2)
3. Spawn Node3
4. Node1 sends a heartbeat that includes clocks for Node1 & Node2
5. Node3 receives the heartbeat. It sees node1 clock is dominating
  because there's Node2 clock. It requests transfer from Node1.
6. Node1 sends transfer ack to Node3
7. Node3 uses `State#extract` to process the transfer payload which
   discards Node2 values.
8. It all starts again from step 4 on the next heartbeat.

This loop between steps 4 and 8 lasts until Node1 permdown period for
Node2 triggers and it doesn't put it to the heartbeat clocks any more.

The solution here is not to include down replicas in the heartbeat
notifications.

This fixes #135

### DIFF
--- a/lib/phoenix/tracker/state.ex
+++ b/lib/phoenix/tracker/state.ex
@@ -68,7 +68,14 @@ defmodule Phoenix.Tracker.State do
   Returns the causal context for the set.
   """
   @spec clocks(t) :: {name, context}
-  def clocks(%State{replica: rep, context: ctx}), do: {rep, ctx}
+  def clocks(%State{replica: rep, context: ctx} = state) do
+    # Exclude down replicas from clocks as they are also not included in
+    # deltas. Otherwise if this node knows of a down node X in permdown grace
+    # period, another node Y which came up after X went down will keep
+    # requesting full state from this node as the clock of Y will be dominated
+    # by the clock of this node.
+    {rep, Map.drop(ctx, down_replicas(state))}
+  end
 
   @doc """
   Adds a new element to the set.
--- a/test/phoenix/tracker/shard_replication_test.exs
+++ b/test/phoenix/tracker/shard_replication_test.exs
@@ -8,11 +8,16 @@ defmodule Phoenix.Tracker.ShardReplicationTest do
 
   setup config do
     tracker = config.test
-    {:ok, shard_pid} = start_shard(name: tracker)
-    {:ok, topic: to_string(config.test),
+
+    tracker_opts = config |> Map.get(:tracker_opts, []) |> Keyword.put_new(:name, tracker)
+    {:ok, shard_pid} = start_shard(tracker_opts)
+
+    {:ok,
+     topic: to_string(tracker),
      shard: shard_name(tracker),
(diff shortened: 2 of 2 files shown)


lib/phoenix/tracker/state.ex
```diff
@@ -68,7 +68,14 @@ defmodule Phoenix.Tracker.State do
   Returns the causal context for the set.
   """
   @spec clocks(t) :: {name, context}
-  def clocks(%State{replica: rep, context: ctx}), do: {rep, ctx}
+  def clocks(%State{replica: rep, context: ctx, replicas: replicas} = state) do
+    down_replicas =
+      replicas
```
urmastalimaa: This can be simplified to 
```elixir
down_replicas = for {key, state} <- replicas, state == :down, do: key
```

(I worked with @indrekj together in investigating and fixing this)

test/phoenix/tracker/shard_replication_test.exs
```diff
@@ -5,14 +5,26 @@ defmodule Phoenix.Tracker.ShardReplicationTest do
...
-     tracker: tracker}
+
+    conf = [
+      topic: to_string(tracker),
+      shard: shard_name(tracker),
+      tracker: tracker
+    ]
+
+    conf =
+      if config[:primary] != false do
```
urmastalimaa: This is actually needlessly complex, 
we can configure extra tracker opts via a tag,
so instead of `primary: false`, we'd have `tracker_opts: [permdown_period: 2000]` as a tag 
and in the test replace node1 with primary and node3 with node1.

chrismccord:
I think this is the way to go but it is also worth noting that a pod being discarded should probably call `graceful_permdown` as part of its shutdown sequence `https://github.com/phoenixframework/phoenix_pubsub/blob/master/lib/phoenix/tracker.ex#L264-L273`. Thanks for digging into this! ❤️❤️❤️🐥🔥

PR #139, opened 2020-04-18
docs(Adapter): include variable names in callbacks

Gazler:
This ensures they are named in exdoc, being called `node_name` instead
of `arg2`, etc.


chrismccord:
❤️❤️❤️🐥🔥

PR #143, opened 2020-05-01
PG2 publish parallel

burningbit:
- Allows registry to publish parallel to processes.


josevalim:
This was how it was implemented previously but we noticed that it actually slowed things down. So we don’t recommend going down this route.

PR #156, opened 2021-04-19
Use Github Actions instead of TravisCI

dolfinus:
Like https://github.com/phoenixframework/phoenix_html/pull/334, but for phoenix_pubsub.
Like https://github.com/phoenixframework/phoenix_pubsub/pull/150 but updated, with only min and max Elixir/OTP versions, and also format and unused dependencies check enabled.


lib/phoenix/pubsub/application.ex
```diff
@@ -15,4 +15,4 @@ defmodule Phoenix.PubSub.Application do
       []
```
dolfinus, line 15: Here and below source code was reformatted with `mix format`

dolfinus:
@josevalim Hi, could you please take a look?

PR #157, opened 2021-05-29
Note that :pg vs :pg2 is checked at compile time

axelson:
This bit me in a project that I'm working on


josevalim:
This shouldn't matter because Elixir compiles a new version if the OTP version changes. :) And now it is outdated anyway. Sorry for late reply.  :)

PR #158, opened 2021-06-17
Add helpful moduledoc to Phoenix.PubSub.Cluster

cjfreeze:
This module has been helpful to me as a guide for writing similar distributed test helpers in several projects, but I've run into issues related to `epmd` not running in the past and forgotten that was the problem/solution. Maybe I'm not alone? Either way, a helpful moduledoc would prevent that in the future!


josevalim:

PR #159, opened 2021-06-30
add: subscribe_uniq/3

mad42:
I've recently encountered edge cases when I subscribe to an already subscribed topic, multiplying received events.
`subscribe_uniq/3` prevents this from happening by looking up in the Registry for `pubsub, topic` before registering it.


chrismccord:
This will only allow a single subscriber *for any process*, so this is not what you want. Note the docs on duplicate subscriptions:
https://github.com/phoenixframework/phoenix_pubsub/blob/master/lib/phoenix/pubsub.ex#L98-L104

The best solution is to avoid duplicating the subscribe in the caller, such as checking your state. Thanks!

PR #165, opened 2021-12-29
Add `Phoenix.PubSub.is_running?/0`

type1fool:
# What is this?

`Phoenix.PubSub.is_started?()` gets the list of registered processes and checks if the provided `pubsub` is in the list.

# Why is this useful?

I am sinfully using PubSub in a GenServer module before PubSub is started. For this, I will repent later, but I needed this function and it seems useful. I couldn't find a way to check PubSub's status without crashing the app (using Supervisor, Application, Registry functions).

Some of my module's functions broadcast when they're called, but the app crashes if PubSub hasn't been started. So, I needed a way to conditionally broadcast messages only if PubSub was running.

# For example

As a temporary solution, I have added a helper to my Application module.

## MyApp.Application

```elixir
def is_pubsub_started?(pubsub) do
  Enum.member?(Process.registered(), pubsub)
end
```

\* The function name is `is_started?` in this PR.

## MyApp.Users

```elixir
def add_user(user) do
  ...
  if MyApp.Application.is_pubsub_started?(MyApp.PubSub) do
    PubSub.broadcast(MyApp.PubSub, @topic, {:user_added, user})
  end
end
```

# Ideal Usage

```elixir
def add_user(user) do
  ...
  if PubSub.is_started?(MyApp.PubSub) do
    PubSub.broadcast(MyApp.PubSub, @topic, {:user_added, user})
  end
end
```

# Feedback

I am open to feedback, even if this feature is deemed out of scope for PubSub. Yes, I will move PubSub out of my User module, I promise. 😇

Thank you for maintaining this. 🫂

### DIFF
--- a/lib/phoenix/pubsub.ex
+++ b/lib/phoenix/pubsub.ex
@@ -289,7 +289,7 @@ defmodule Phoenix.PubSub do
   end
 
   @doc """
-  Returns `true` if Phoenix PubSub has been started, otherwise it returns false.
+  Returns `true` if Phoenix PubSub has been started, otherwise it returns `false`.
   """
   def is_running? do
     started_applications = Application.started_applications() |> Enum.map(&elem(&1, 0))
--- a/lib/phoenix/pubsub.ex
+++ b/lib/phoenix/pubsub.ex
@@ -34,7 +34,7 @@ defmodule Phoenix.PubSub do
       It supports a `:pool_size` option to be given alongside
       the name, defaults to `1`. Note the `:pool_size` must
       be the same throughout the cluster, therefore don't
-      configure the pool size based on `System.schedulers_online/1`, 
+      configure the pool size based on `System.schedulers_online/1`,
       especially if you are using machines with different specs.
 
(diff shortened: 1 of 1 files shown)


type1fool:
Somehow this worked yesterday, but not today. I'm closing the PR until I can find an alternative solution.

type1fool:
I'm reopening this after finding a simpler and more reliable solution thanks to `Process.registered/0`. While it would be easy enough to just use the body of `is_started/1` where I need it, this was a tricky problem to solve. I imagine this new function would be convenient for folks in the future.

type1fool:
Today, it was pointed out that `Phoenix.PubSub` could be moved in `application.ex`'s children list. When I have it started before the modules that use it, there is no issue.

So, I'll leave this open just in case the new function is still useful.

chrismccord:
> Today, it was pointed out that `Phoenix.PubSub` could be moved in `application.ex`'s children list. When I have it started before the modules that use it, there is no issue.

Yes, this is the correct approach. Thanks!

PR #167, opened 2022-03-22
Make it clear that `PG2` is not related to Postgres

thbar:
Recently working on a multi-node app which is not yet an Elixir cluster, I somehow assumed that the `PG2` adapter would help me broadcast from one node to another, and it took a bit of time & Slack exchanges to realise that `PG2` is in no way related to Postgres.

The confusion probably stems from the fact that `pg` is, for a number of popular stacks (Ruby, Javascript), the name of the client library for Postgres.

### Remaining tasks

- [ ] Add a blurb here https://hexdocs.pm/phoenix_pubsub/Phoenix.PubSub.html#module-adapters too

### DIFF
--- a/lib/phoenix/pubsub/pg2.ex
+++ b/lib/phoenix/pubsub/pg2.ex
@@ -3,6 +3,10 @@ defmodule Phoenix.PubSub.PG2 do
   Phoenix PubSub adapter based on `:pg`/`:pg2`.
 
   It runs on Distributed Erlang and is the default adapter.
+  
+  Please note that `pg` stands for "Process Group" (an Erlang module).
+
+  This PubSub adapter is in no way related to Postgres, the database.
   """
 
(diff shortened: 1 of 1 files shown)


chrismccord:
> I somehow assumed that the `PG2` adapter would help me broadcast from one node to another

This is the correct assumption, pg2 is a distributed process group :)

I'd prefer to wait to see if this comes up again in the future. Thanks!

thbar:
Sorry, my sentence is missing a large part: I somehow assumed that the PG2 adapter would help me broadcast from one node to another "without having a proper Elixir cluster" (we're using a PaaS with Docker deploy, 2 nodes which are only communicating via Postgres). This made it quite confusing.

> I'd prefer to wait to see if this comes up again in the future. Thanks!

Good call, thanks for the quick answer!

Issue #168, opened 2022-04-05
v2.1 and v2.0 are incompatible


nmbrone:
Hi all.

Not sure this will be considered an issue.

We have a use-case for the library where we have service A that broadcasts some payload and service B that consumes it via PubSub (A and B are in a cluster, both running PubSub v2.0).

In v2.1 some changes related to the process grouping were made in the `Phoenix.PubSub.PG2` adapter. Now, if one of the services named above is updated to v2.1 independently, communication between services is not possible anymore.

PR #169, opened 2022-04-05
Rename first adapter in the pool for backwards compatability

Gazler:
Pubsub v2.0 expects there to be a single process for the process group
named `Elixir.my_pubsub.Adapter`.

v2.1 added pools, which means that the processes are supervised with a
supervisor called `Elixir.my_pubsub.Adapter` along with n processes
named `Elixir.my_pubsub.Adapter_n` where n is the pool size.

This commit changes the supervisor name to
`Elixir.my_pubsub.Adapter_supervisor`
and renames pool item 1 to be `Elixir.my_pubsub.Adapter`. This will
allow backwards compatability with v2.0 when the pool size is set to 1.


josevalim (approved):

chrismccord:
Thanks. I will cut a new 2.1.1. This also closes #168 ❤️❤️❤️🐥🔥

PR #172, opened 2023-05-24
Update ETS atomically on tracker update

jonatanklosko:
Tracker update is handled as leave + join and ETS updates are applied in this notion, that is, a delete followed by an insert. Since tracker lookup reads ETS on the caller side, it can happen concurrently to tracker update, so there is a race condition when the ETS read happens between said delete and insert.

The "values" ETS table (which keeps the metadata) is an :ordered_set, which means that we can do an atomic update just by inserting the new value, without prior delete. So I made two changes:

  * for the local case I added `State.leave_join`, which works just like `State.leave` + `State.join`, but without the unnecessary delete
  * for the remote case I updated `State.merge` to gather desired inserts/deletes to ETS, do a simple diff and only then apply the changes (which makes an entry update atomic)

The changes only affect applying changes to the internal ETS storage, the update is reported as leave + join in exactly the same way.

Context: in Livebook we use tracker for registering notebook session processes and we also store basic session information necessary for listing, which gets updated. We have a lot of tests that create and lookup sessions and we started to observe more and more intermittent test failures that basically go: create session -> lookup session -> session not found. Running tests in a loop I could reproduce this behaviour and tracked it down to this race condition, with this PR it no longer happens.

### DIFF
--- a/lib/phoenix/tracker/state.ex
+++ b/lib/phoenix/tracker/state.ex
@@ -84,6 +84,27 @@ defmodule Phoenix.Tracker.State do
   end
 
   @doc """
+  Updates an element via leave and join.
+
+  Atomically updates ETS local entry.
+  """
+  @spec leave_join(t, pid, topic, key, meta) :: t
+  def leave_join(%State{delta: delta} = state, pid, topic, key, meta) do
+    # Produce remove-like delta
+    [{{^topic, ^pid, ^key}, _meta, tag}] = :ets.lookup(state.values, {topic, pid, key})
+    pruned_clouds = delete_tag(state.clouds, tag)
+    new_delta = remove_delta_tag(state.delta, tag)
+    state = bump_clock(%State{state | clouds: pruned_clouds, delta: new_delta})
+
+    # Update ETS entry and produce add-like delta
+    state = bump_clock(state)
+    tag = tag(state)
+    true = :ets.insert(state.values, {{topic, pid, key}, meta, tag})
+    new_delta = %State{delta | values: Map.put(delta.values, tag, {pid, topic, key, meta})}
+    %State{state | delta: new_delta}
+  end
+
+  @doc """
   Removes an element from the set.
   """
   @spec leave(t, pid, topic, key) :: t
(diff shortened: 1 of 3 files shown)


lib/phoenix/tracker/state.ex
```diff
@@ -83,6 +83,27 @@ defmodule Phoenix.Tracker.State do
...
+  @spec leave_join(t, pid, topic, key, meta) :: t
+  def leave_join(%State{delta: delta} = state, pid, topic, key, meta) do
+    # Produce remove-like delta
+    [{{^topic, ^pid, ^key}, _meta, tag}] = :ets.lookup(state.values, {topic, pid, key})
+    pruned_clouds = delete_tag(state.clouds, tag)
+    new_delta = remove_delta_tag(state.delta, tag)
+    state = bump_clock(%State{state | clouds: pruned_clouds, delta: new_delta})
+
+    # Update ETS entry and produce add-like delta
+    state = bump_clock(state)
```
jonatanklosko, line 100: Not sure if we need to bump the clock twice, I just mirrored the exact behaviour of leave + join.

chrismccord:
❤️❤️❤️🐥🔥

chrismccord:
Released in v2.1.2. Thank you so much!

jonatanklosko:
Beautiful, thanks! <3

arjan:
With this change I am seeing issues where the presence ETS values tables on different nodes are drifting out of sync. It seems that presence entries from processeses on other nodes are not always cleaned up. This forced me to roll back to 2.1.1 for the time being. I will see if I can come up with a reproducible case.

jonatanklosko:
@arjan are you comparing the ETS rows, or the result of `Presence.list`? Also, do they differ in excess processes or the metadata itself?

I went through the changes a couple times and so far haven't pinpointed anything that would introduce any meaningful difference.

arjan:
I noticed it because our stats use the output of `Presence.list` and the counts were way off yesterday morning. Hopefully I can reproduce it today.

Issue #175, opened 2023-06-14
Presence list keep growing when using Presence.update


arjan:
It seems #172 (released in 2.1.2) seems to have introduced a bug where the list of presences keeps growing when a process uses Presence.update.

## How to reproduce

The following gist is a test script that starts a PubSub server and a Phoenix presence tracker.

https://gist.github.com/arjan/6702b4b9fd6b97e9b1a355313ed1620b

Periodically, genserver processes are started that do `Presence.track` and, a bit later, `Presence.update`, and another bit later stop again.

A poller process runs that shows the number of running (local) processes, the number of items in the presence list, and the number of items in the presence list split out by node.

Run the test script in 2 terminals with:

```
▶ elixir --name a@127.0.0.1 --cookie hello ./presence_bug.exs
▶ elixir --name b@127.0.0.1 --cookie hello ./presence_bug.exs
```

When running on 2.1.1, the presence count is stable, settling at about ~100 processes per node.

```
10:30:43.465 [warning] Local processes: 106 - Presence count: 178 - Per node counts: ["a@127.0.0.1": 106, "b@127.0.0.1": 110]
```

When running on 2.1.2, the presence count keeps increasing, because the presences for the non-local nodes are not cleaned up, the per-node counts for the non-local node keeps increasing, eg:

```
10:29:55.630 [warning] Local processes: 100 - Presence count: 355 - Per node counts: ["a@127.0.0.1": 322, "b@127.0.0.1": 100]
```

MattiasOlla:
Thanks for reporting, we also had issues with this. It also led to our application consuming more than twice as much more memory as before updating:

![image](https://github.com/phoenixframework/phoenix_pubsub/assets/17568632/a66d2b94-a4a9-42d5-9832-42b8a6a00fd3)

jonatanklosko:
@arjan thanks for the great reproduction! I'm looking into this.

chrismccord:
Fixed in 2.1.3

PR #176, opened 2023-06-14
Fix leftover entries after distributed tracker update

jonatanklosko:
Closes https://github.com/phoenixframework/phoenix_pubsub/issues/175.

### DIFF
--- a/lib/phoenix/tracker/state.ex
+++ b/lib/phoenix/tracker/state.ex
@@ -89,7 +89,7 @@ defmodule Phoenix.Tracker.State do
   Atomically updates ETS local entry.
   """
   @spec leave_join(t, pid, topic, key, meta) :: t
-  def leave_join(%State{delta: delta} = state, pid, topic, key, meta) do
+  def leave_join(state, pid, topic, key, meta) do
     # Produce remove-like delta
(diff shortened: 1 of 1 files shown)


lib/phoenix/tracker/state.ex
```diff
@@ -89,7 +89,7 @@ defmodule Phoenix.Tracker.State do
   Atomically updates ETS local entry.
   """
   @spec leave_join(t, pid, topic, key, meta) :: t
-  def leave_join(%State{delta: delta} = state, pid, topic, key, meta) do
+  def leave_join(state, pid, topic, key, meta) do
```
jonatanklosko, line 92: When merging leave and join I kept the pattern match, so the join part operated on delta from before leave. Totally my bad, sorry!

jonatanklosko:
@arjan I believe this is it, sorry for the inconvenience and thanks again for the reproduction.

arjan:
Yes looks like this fixes it, thanks!

chrismccord:
❤️❤️❤️🐥🔥

chrismccord:
Released with 2.1.3

arjan:
Amazing, thanks a lot!

PR #177, opened 2023-07-11
Add simple DNS clustering

chrismccord:


lib/phoenix/pubsub/dns_cluster.ex
```diff
@@ -0,0 +1,139 @@
...
+    GenServer.start_link(__MODULE__, opts, name: Keyword.get(opts, :name, __MODULE__))
+  end
+
+  @impl true
+  def init(opts) do
+    resolver = Keyword.get(opts, :resolver, Resolver)
+
+    # discover nodes on init so we're connected before downstream children start
+    state =
+      do_discovery(%{
```
chrismccord, line 71: we can later introduce a `:sync_init` option or similar, but I feel like it's best to do synchronous discovery on init w/ timeout, to lessen races on downstream children starting without full mesh connections

lib/phoenix/pubsub/dns_cluster.ex
```diff
@@ -0,0 +1,139 @@
...
+              log(state, "#{node()} connected to #{new_name}")
+            end
+          end,
+          max_concurrency: length(ips),
+          timeout: timeout
+        )
+        |> Enum.to_list()
+      end)
+
+    case Task.yield(task, timeout) || Task.shutdown(task, :brutal_kill) do
```
josevalim, line 117: If you kill the task you will crash yourself.

lib/phoenix/pubsub/pg2.ex
```diff
@@ -83,6 +98,16 @@ defmodule Phoenix.PubSub.PG2 do
         Supervisor.child_spec({Phoenix.PubSub.PG2Worker, {name, group}}, id: group)
       end
 
+    children =
+      case System.get_env("PHX_CLUSTER_DNS") do
```
josevalim, line 102: We really shouldn't be relying on env vars down here. Also, the cluster name is not related to PubSub, so I would say this functionality is out.

My suggestion is to add `Phoenix.PubSub.DNSCluster` as a service and have your Phoenix endpoint start it based on a config:

    config :my_app, MyApp.Endpoint, dns_cluster: true

lib/phoenix/pubsub/dns_cluster.ex
```diff
@@ -0,0 +1,139 @@
...
+
+  See the `start_link/1` docs for all available options.
+
+  If you require more advanced clustering options and strategies, see the
+  [libcluster](https://hexdocs.pm/libcluster) library.
+  """
+  use GenServer
+  require Logger
+
+  defmodule Resolver do
```
josevalim, line 30: Why is it a separate module? 🤔

chrismccord: just for testing. Alternative is doc false'ing everything on the public DNSCliter

lib/phoenix/pubsub/pg2.ex
chrismccord: It's not "related", but also the pg adapter is not useful if you aren't clustered, so it's still pretty well related. I'm fine with the endpoint starting it as well, but we have a problem in that the pubsub adapter is started above the endpoint for good reasons, so we'd be starting pubsub before clustering, which seems backwards, so I'm not sure?

chrismccord: So to put another way, I'm convinced we need a knob free way to get out of the box clustering, but if we don't put in the pubsub tree we don't have any other option to start it in the correct order, it short of adding it explicitly adding it to the children in `phx.new`

chrismccord:
Closing in favor of separate library https://github.com/phoenixframework/dns_cluster

PR #178, opened 2023-08-11
Fix type-spec for Phoenix.Tracker.get_by_key

ausimian:


ausimian:
`Shard.get_by_key` also has the same error so will fix that as well

chrismccord:
❤️❤️❤️🐥🔥

ausimian:
hey thanks for responding so quickly! is there any chance this could be released as 2.1.4 at some point? The invalid spec results in type errors that propagate around my code base and pulling the dep from github is not possible in my work environment 😢

PR #184, opened 2024-05-24
Only allow update if pid is on current node

schrockwell:
To the best of my knowledge, `Phoenix.Tracker.update/5` should not be used to update presence metadata for processes that are on a separate node. If you do this, Bad Things will happen. Ask me how I know. 

So, this PR adds a safeguard against this.

Returning `{:error, :nopresence}` might also be acceptable here for backwards-compatibility, at the cost of an error reason that isn't really accurate.

I did my best to implement the test case, but I couldn't follow 100% what is going on, so there may be a better way to set it up.

### DIFF
--- a/lib/phoenix/tracker/shard.ex
+++ b/lib/phoenix/tracker/shard.ex
@@ -531,6 +531,9 @@ defmodule Phoenix.Tracker.Shard do
     """
   end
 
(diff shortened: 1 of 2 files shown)


josevalim (approved):

davydog187:
@josevalim is this PR ok to merge?

josevalim:
If I didn't merge it is because I didn't feel comfortable doing so before. :) We need more eyes on it!

PR #190, opened 2024-09-26
Add ability to replicate state immediately during normal shutdown

nathanl:
Add ability to replicate state immediately during normal shutdown via new option

### DIFF
--- a/lib/phoenix/tracker.ex
+++ b/lib/phoenix/tracker.ex
@@ -72,12 +72,12 @@ defmodule Phoenix.Tracker do
 
   ## Application Shutdown
 
-  When a tracker shuts down, the other nodes do not assume it has been gone
+  When a tracker shuts down, the other nodes do not assume it is gone
   for good. After all, in a distributed system, it is impossible to know if something
   is just temporarily unavailable or if it has crashed.
 
   For this reason, when you call `System.stop()` or the Erlang VM receives a
-  `SIGTERM` - any presences that the local tracker instance has will continue to
+  `SIGTERM`, any presences that the local tracker instance has will continue to
   be seen as present by other trackers in the cluster until the `:down_period`
   for the instance has passed.
 
@@ -292,7 +292,7 @@ defmodule Phoenix.Tracker do
       (30s down detection). Note: This must be at least 2x the `broadcast_period`.
     * `permdown_on_shutdown` - boolean; whether to immediately call
       `graceful_permdown/1` on the tracker during a graceful shutdown. See
-      'Application Shutdown' section. You can only safely set this point if `Phoenix.Tracker`
+      'Application Shutdown' section. You can only safely set this if `Phoenix.Tracker`
       is mounted at the root of your supervision tree and the strategy is `:one_for_one`.
       Default `false`.
     * `:permdown_period` - The interval in milliseconds to flag a replica
--- a/lib/phoenix/tracker.ex
+++ b/lib/phoenix/tracker.ex
@@ -292,7 +292,9 @@ defmodule Phoenix.Tracker do
       (30s down detection). Note: This must be at least 2x the `broadcast_period`.
     * `permdown_on_shutdown` - boolean; whether to immediately call
       `graceful_permdown/1` on the tracker during a graceful shutdown. See
-      'Application Shutdown' section. Default `false`.
+      'Application Shutdown' section. You can only safely set this point if `Phoenix.Tracker`
+      is mounted at the root of your supervision tree and the strategy is `:one_for_one`.
+      Default `false`.
     * `:permdown_period` - The interval in milliseconds to flag a replica
       as permanently down, and discard its state.
       Note: This must be at least greater than the `down_period`.
--- a/lib/phoenix/tracker.ex
+++ b/lib/phoenix/tracker.ex
@@ -72,12 +72,14 @@ defmodule Phoenix.Tracker do
 
   ## Application Shutdown
 
-  A tracker does not automatically replicate its state across the cluster as it
-  shuts down. This means that your supervision tree shuts down normally - as it
-  does when you call `System.stop()` or the BEAM receives a `SIGTERM` - any
-  presences that the local tracker instance has will continue to be seen as
-  present by other trackers in the cluster until the `:down_period` for the
-  instance has passed.
+  When a tracker shuts down, the other nodes do not assume it has been gone
+  for good. After all, in a distributed system, it is impossible to know if something
+  is just temporarily unavailable or if it has crashed.
+
+  For this reason, when you call `System.stop()` or the Erlang VM receives a
+  `SIGTERM` - any presences that the local tracker instance has will continue to
+  be seen as present by other trackers in the cluster until the `:down_period`
+  for the instance has passed.
 
   If you want a normal shutdown to immediately cause other nodes to see that
   tracker's presences as leaving, pass `permdown_on_shutdown: true`. On the
--- a/lib/phoenix/tracker.ex
+++ b/lib/phoenix/tracker.ex
@@ -62,13 +62,34 @@ defmodule Phoenix.Tracker do
   An optional `handle_info/2` callback may also be invoked to handle
   application specific messages within your tracker.
 
-  ## Special Considerations
+  ## Stability and Performance Considerations
 
   Operations within `handle_diff/2` happen *in the tracker server's context*.
   Therefore, blocking operations should be avoided when possible, and offloaded
   to a supervised task when required. Also, a crash in the `handle_diff/2` will
   crash the tracker server, so operations that may crash the server should be
   offloaded with a `Task.Supervisor` spawned process.
+
+  ## Application Shutdown
+
+  A tracker does not automatically replicate its state across the cluster as it
+  shuts down. This means that your supervision tree shuts down normally - as it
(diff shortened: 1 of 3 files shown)


lib/phoenix/tracker.ex
```diff
@@ -62,13 +62,48 @@ defmodule Phoenix.Tracker do
...
+          Process.flag(:trap_exit, true)
+          {:ok, init_arg}
+        end
+      
+        @impl GenServer
+        def terminate(_reason, _state) do
+          Phoenix.Tracker.graceful_permdown(MyAppWeb.Tracker)
+          :ok
+        end
+      end
```
josevalim, line 106: The problem with this approach is that it does not distinguish between a clean terminate and a non-clean one. Maybe we should check the reason? But I am not sure if the supervisor will pass the relevant information either.

nathanl: Are you referring to the ignored `_reason`? My thinking was that since this GenServer doesn't do anything except wait for termination, it should only terminate when the supervision tree is shutting down, at which point the Tracker is going to be the next thing to go down. When might something go wrong here?

nathanl: If the BEAM crashes, this `terminate` obviously won't run, but you'd be no worse off than without it; you'd be back to having the other nodes wait the `:down_period`.

josevalim: > it should only terminate when the supervision tree is shutting down

Yes, but a supervision tree may also shut down and be restarted. Or this is being shutdown because the restart strategy is `:rest_for_one` or `:one_for_all`. So this is assuming this is in the root supervision tree AND the strategy is `:one_for_one`. We should probably make these assumptions clearer?

josevalim: I also wonder then why we don't add this as an option to the tracker itself:

    perm_down_on_shutdown: ...

But maybe there are other implications there.

josevalim: I also wonder if we should add a `{Task.Shutdown, fn -> ... end}` to Elixir that makes it easier to run something only on shutdown.

nathanl: > Yes, but a supervision tree may also shut down and be restarted.
Even if that's the case, if the tracker is about to shut down, it's about to lose its state, right? Anything that it has tracked, it will no longer know about until they call `.track` again after the restart. If it's state is about to be lost, we can either have the other nodes find out about that or be out of sync?

As to your ideas for building this in somehow, I would love that. I was very surprised that this didn't get handled automatically, so part of the reason for opening this PR was to confirm/deny that this was the right approach.

josevalim: > Even if that's the case, if the tracker is about to shut down, it's about to lose its state, right?

Yes, but in Phoenix.Presence cases, the clients will reconnect, and not issuing a perm down means that clients don't immediately get a bunch of notifications that thousands or dozens of thousands of users left. They will see new metadata joining and then leaving, but with a time between them. Instead of leaves and joins.

In any case, I think we can have a `permdown_on_shutdown` option which can be set to true and bake this in. :)

nathanl: Do you want some modified version of this documentation in the meantime? Do you want a PR for `perm_on_shutdown`? (If so, I could use some guidance on where to implement that.)

josevalim: >  Do you want a PR for perm_on_shutdown? (If so, I could use some guidance on where to implement that.)

Yes, please! `Phoenix.Tracker` is a supervisor, you need to add a child similar to the one you documented here. :)

lib/phoenix/tracker.ex
```diff
@@ -62,13 +62,34 @@ defmodule Phoenix.Tracker do
...
+  tracker's presences as leaving, pass `permdown_on_shutdown: true`. On the
+  other hand, if you are using `Phoenix.Presence` for clients which will
+  immediately attempt to connect to a new node, it may be preferable to use
+  `permdown_on_shutdown: false`, allowing the disconnected clients time to
+  reconnect before removing their old presences, to avoid overwhelming clients
+  with notifications that many users left and immediately rejoined.
+
+  If the application crashes or is halted non-gracefully (for instance, with a
+  `SIGKILL` or a `Ctrl+C` in `iex`), other nodes will still have to wait the
+  `:down_period` to notice that the tracker's presences are gone.
```
josevalim, line 92:
```suggestion
  When a tracker shuts down, the other nodes do not assume it has been gone
  for good. After all, in a distributed system, it is impossible to know if something
  is just temporarily unavailable or if it has crashed.

  For this reason, when you call `System.stop()` or the Erlang VM receives a
  `SIGTERM` - any presences that the local tracker instance has will continue to
  be seen as present by other trackers in the cluster until the `:down_period`
  for the instance has passed.

  If you want a normal shutdown to immediately cause other nodes to see that
  tracker's presences as leaving, pass `permdown_on_shutdown: true`. On the
  other hand, if you are using `Phoenix.Presence` for clients which will
  immediately attempt to connect to a new node, it may be preferable to use
  `permdown_on_shutdown: false`, allowing the disconnected clients time to
  reconnect before removing their old presences, to avoid overwhelming clients
  with notifications that many users left and immediately rejoined.

  If the application crashes or is halted non-gracefully (for instance, with a
  `SIGKILL` or a `Ctrl+C` in `iex`), other nodes will still have to wait the
  `:down_period` to notice that the tracker's presences are gone.
```

lib/phoenix/tracker.ex
```diff
@@ -267,6 +288,9 @@ defmodule Phoenix.Tracker do
     * `:down_period` - The interval in milliseconds to flag a replica
       as temporarily down. Default `broadcast_period * max_silent_periods * 2`
       (30s down detection). Note: This must be at least 2x the `broadcast_period`.
+    * `permdown_on_shutdown` - boolean; whether to immediately call
+      `graceful_permdown/1` on the tracker during a graceful shutdown. See
+      'Application Shutdown' section. Default `false`.
```
josevalim, line 293:
```suggestion
      'Application Shutdown' section. You can only safely set this point if `Phoenix.Tracker`
      is mounted at the root of your supervision tree and the strategy is `:one_for_one`.
      Default `false`.
```

lib/phoenix/tracker.ex
```diff
@@ -301,13 +326,24 @@ defmodule Phoenix.Tracker do
         }
       end
 
+    children = if permdown_on_shutdown do
+      shards ++ [
+        %{
+          id: :shutdown_handler,
+          start: {Phoenix.Tracker.ShutdownHandler, :start_link, [[tracker: tracker]]}
```
josevalim, line 333:
```suggestion
          start: {Phoenix.Tracker.ShutdownHandler, :start_link, [tracker]}
```

lib/phoenix/tracker/shutdown_handler.ex
```diff
@@ -0,0 +1,21 @@
...
+    tracker = Keyword.fetch!(opts, :tracker)
+    Process.flag(:trap_exit, true)
+    {:ok, %{tracker: tracker}}
+  end
+
+  @impl GenServer
+  def terminate(_reason, state) do
+    Phoenix.Tracker.graceful_permdown(state.tracker)
+    :ok
+  end
```
josevalim, line 20:
```suggestion
  def start_link(tracker) do
    GenServer.start_link(__MODULE__, tracker, name: __MODULE__)
  end

  @impl GenServer
  def init(tracker) do
    Process.flag(:trap_exit, true)
    {:ok, tracker}
  end

  @impl GenServer
  def terminate(_reason, tracker) do
    Phoenix.Tracker.graceful_permdown(tracker)
    :ok
  end
```

lib/phoenix/tracker.ex
```diff
@@ -62,13 +62,36 @@ defmodule Phoenix.Tracker do
...
 
   Operations within `handle_diff/2` happen *in the tracker server's context*.
   Therefore, blocking operations should be avoided when possible, and offloaded
   to a supervised task when required. Also, a crash in the `handle_diff/2` will
   crash the tracker server, so operations that may crash the server should be
   offloaded with a `Task.Supervisor` spawned process.
+
+  ## Application Shutdown
+
+  When a tracker shuts down, the other nodes do not assume it has been gone
```
nathanl, line 75:
```suggestion
  When a tracker shuts down, the other nodes do not assume it is gone
```

lib/phoenix/tracker.ex
```diff
@@ -267,6 +290,11 @@ defmodule Phoenix.Tracker do
     * `:down_period` - The interval in milliseconds to flag a replica
       as temporarily down. Default `broadcast_period * max_silent_periods * 2`
       (30s down detection). Note: This must be at least 2x the `broadcast_period`.
+    * `permdown_on_shutdown` - boolean; whether to immediately call
+      `graceful_permdown/1` on the tracker during a graceful shutdown. See
+      'Application Shutdown' section. You can only safely set this point if `Phoenix.Tracker`
```
nathanl, line 295:
```suggestion
      'Application Shutdown' section. You can only safely set this if `Phoenix.Tracker`
```

chrismccord:
❤️❤️❤️🐥🔥

nathanl:
Hat tip to Jason Axelson for chiming in that, like me, he was doing something like this in his app: https://felt.com/blog/startup-and-shutdown-for-phoenix-applications

PR #192, opened 2024-10-14
Added unsubscribe function that matches on metadata

m0rt3nlund:
This allows multiple subscriptions to the same topic and later the ability to remove a specific subscription instead of "all" subscriptions for the topic


davydog187:
Hello @m0rt3nlund, I am not a maintainer, but this change looks good to me. Would you be able to add a test?

davydog187:
I added a test on your behalf in #206

PR #199, opened 2025-06-25
Add `Phoenix.PubSub.subscribe_once/3`

nathanl:
Given that the possibility of duplicate subscriptions was already discussed in the docs on `subscribe/3`, I assume there's a performance cost to checking for existing subscriptions that callers may not always want to pay. If so, I'd be happy to document that on `subscribe_once/3`. But it seems useful to have this function; I'm looking at a use case where the number of subscriptions to a topic won't be large but duplicate messages would be problematic.

Currently, any caller who wants to avoid duplicate subscriptions would need to use some kind of workaround, such as:

- Avoid calling `subscribe/3` more than once by tracking subscriptions in their own process state, or by checking `Registry.lookup/2` directly (relying on an implementation detail of PubSub)
- Call `unsubscribe/2` each time before calling `subscribe/3`

### DIFF
--- a/lib/phoenix/pubsub.ex
+++ b/lib/phoenix/pubsub.ex
@@ -191,6 +191,8 @@ defmodule Phoenix.PubSub do
   `Phoenix.PubSub.unsubscribe/2`, all duplicate subscriptions
   will be dropped.
 
+  See `subscribe_once/3` to avoid duplicate subscriptions.
+
   ## Options
 
     * `:metadata` - provides metadata to be attached to this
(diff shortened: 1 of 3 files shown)


lib/phoenix/pubsub.ex
```diff
@@ -208,6 +210,22 @@ defmodule Phoenix.PubSub do
...
 
+  @doc """
+  Subscribes the caller to the PubSub adapter's topic like `subscribe/3` unless
+  the calling process is already subscribed, in which case it returns `{:error,
+  :already_subscribed}`.
+  """
+  @spec subscribe_once(t, topic, keyword) :: :ok | {:error, term}
+  def subscribe_once(pubsub, topic, opts \\ [])
+      when is_atom(pubsub) and is_binary(topic) and is_list(opts) do
+    subscriptions = Registry.lookup(pubsub, topic)
```
davydog187, line 221: Unfortunately, I believe there is a race condition here. In betwen the `Registry` lookup and the `else` path, another process could have subscribed.

The only way this could work is by doing an [insert_new](https://www.erlang.org/doc/apps/stdlib/ets.html#insert_new/2) operation to `:ets`

josevalim: Since you can only subscribe yourself, it cannot be a race condition. However, because this is a duplicate registry, there is basicaly no cheap way to check if you are already subscribed. If you have a topic with 100k entries, then this function will be very expensive because you are bringing a 100k elements list to memory, which is why we don't provide this functionality, it is basically an anti-pattern. So it is more scalable to guarantee code wise you won't subscribe multiple times. :)

josevalim:
Closing per the above. :)

lib/phoenix/pubsub.ex
nathanl: > a topic with 100k entries

Meaning 100k subscribers?

nathanl: Given what you said, I'd like to do a followup docs PR to explain why the caller should guard against multiple subscriptions

josevalim: correct!

josevalim: That would be lovely!!! :heart:

PR #204, opened 2025-12-02
Fix the misattributed root cause of broken tests

zekedou:
Although the CI was fixed by #203, the attribution was incorrect. 

The cause should be the Elixir version - [a feature of registry introduced in Elixir v1.19](https://github.com/elixir-lang/elixir/pull/14654), rather than the Erlang/OTP version.

This PR fixes it.


SteffenDE (approved):
Good call!

josevalim:

PR #205, opened 2026-01-29
chore: run the formatter

davydog187:
Adds a minimal formatter config and runs the formatter on the whole codebase. No other changes have been made.


SteffenDE (approved):

davydog187:
~~if this is to be accepted, i should add the formatter check to CI as well~~

Done

PR #206, opened 2026-01-30
Added unsubscribe function that matches on metadata

davydog187:
Adds a test to #192


josevalim:
We should document the additional argument is a match, since it has some special rules. I would personally add the _match suffix to this function.

davydog187:
@josevalim done

josevalim:

PR #207, opened 2026-02-05
fix: use :peer for cluster tests

davydog187:
This is logging a warning on Elixir 1.19.3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,12 +71,6 @@ jobs:
         env:
           MIX_ENV: test
 
-      - name: Check if formatted
-        run: mix format --check-formatted
-        if: ${{ matrix.lint }}
-        env:
-          MIX_ENV: test
-
       - name: Run tests
         run: |
           epmd -daemon
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,56 +1,83 @@
-name: Tests
+name: CI
 
 on:
-  pull_request:
   push:
-
-env:
-  ACTIONS_RUNNER_FORCED_INTERNAL_NODE_VERSION: node16
-  ACTIONS_RUNNER_FORCE_ACTIONS_NODE_VERSION: node16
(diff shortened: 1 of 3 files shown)


test/support/cluster.ex
```diff
@@ -9,17 +9,23 @@ defmodule Phoenix.PubSub.Cluster do
...
+    cookie = :erlang.get_cookie()
+
+    {:ok, _peer, node} =
+      :peer.start(%{
+        name: node_name(node_host),
+        host: ~c"127.0.0.1",
+        env: [{~c"ERL_AFLAGS", ~c"-setcookie #{cookie}"}]
+      })
+
+    # Ensure bidirectional connection
```
davydog187:
```suggestion
```

SteffenDE:
I guess we can increase the minimum required elixir version

josevalim:
Yes, let's bump both Erlang and Elixir and we can ship it.

davydog187:
@josevalim i copied over the CI workflow from phoenix live view and changed the minimum supported version to OTP 25, which is the earliest that supports `:peer`.

For the library itself, I see no reason to change since this is just a test dependency. Let me know if you want to change, its currently Elixir 1.6

.github/workflows/ci.yml
```diff
@@ -1,56 +1,77 @@
...
+  mix_test:
+    name: mix test (OTP ${{matrix.otp}} | Elixir ${{matrix.elixir}})
 
     strategy:
       fail-fast: false
       matrix:
         include:
-          - image: 1.6.6-erlang-21.3.8.24-debian-buster-20210902-slim
-          - image: 1.19.3-erlang-28.1.1-debian-bullseye-20251103-slim
+          - elixir: 1.15.4
```
josevalim: The earlier to support 25, otherwise please bump to 1.14

```suggestion
          - elixir: 1.13.4
```

josevalim:
If we don't test it on v1.6, then we shouldn't say we support it. We can likely update it to `~> 1.13`.

davydog187:
@josevalim done

josevalim:

PR #209, opened 2026-04-15
Add configurable default dispatcher

twinn:
## Summary

Adds a `:dispatcher` option to `child_spec`/`start_link` that sets the default dispatcher module for all broadcast functions. Defaults to `Phoenix.PubSub` (preserving existing behavior). Can be overridden per-call by passing a dispatcher to `broadcast/4` and friends.

```elixir
{Phoenix.PubSub, name: :my_pubsub, dispatcher: MyApp.Dispatcher}
```

## Motivation

The custom dispatching feature is powerful, but in practice it's difficult to use as a default for an application. The current options are:

1. Pass the dispatcher explicitly on every `broadcast` call
2. Create a wrapper module that injects the dispatcher

Both add boilerplate for what is conceptually a per-PubSub configuration. The wrapper approach is particularly fragile. Any code that calls `Phoenix.PubSub.broadcast/3` directly instead of going through the wrapper bypasses the custom dispatcher entirely. In a growing codebase with multiple contributors, this is easy to miss.

The use case I ran into was filtering broadcasts in test environments, scoping delivery so that concurrent async tests sharing a topic don't leak messages to each other. A configurable default makes this a one-line change at startup rather than a module-level abstraction that every callsite must remember to use.

## Implementation

The internal meta `:pubsub` tuple is extended from `{adapter, name}` to `{adapter, name, dispatcher}`. Single `Registry.meta` lookup per broadcast call, same as before. No additional lookups.

## Changes

- `Phoenix.PubSub.Supervisor` accepts `:dispatcher` option, stores it in the meta tuple
- All broadcast functions default dispatcher to `nil`, read the configured default from the meta tuple, and fall back to it when no explicit dispatcher is passed
- 6 new tests covering default dispatcher, per-call override, and all broadcast variants

### DIFF
--- a/lib/phoenix/pubsub.ex
+++ b/lib/phoenix/pubsub.ex
@@ -171,6 +171,9 @@ defmodule Phoenix.PubSub do
     * `:broadcast_pool_size` - number of pubsub partitions used for broadcasting messages
       (defaults to `:pool_size`). This option is used during pool size migrations to ensure
       no messages are lost. See the "Safe Pool Size Migration" section in the module documentation.
+    * `:dispatcher` - the default dispatcher module for broadcasts
+      (defaults to `Phoenix.PubSub`). Can be overridden per-call by
+      passing a dispatcher to `broadcast/4` and friends.
 
   """
   @spec child_spec(keyword) :: Supervisor.child_spec()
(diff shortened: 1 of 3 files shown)


josevalim:

PR #210, opened 2026-04-28
Add configurable :group_by option for the underlying Registry

studzien:
## Summary

Adds a `:group_by` option to `Phoenix.PubSub`, forwarded to the underlying `Registry`. Defaults to `:pid` (no behavior change). Accepts `:key` or `pid`.

```elixir
{Phoenix.PubSub, name: :my_pubsub, group_by: :pid}
```

## Motivation

`{:duplicate, :key}`, added in Registry [elixir-lang/elixir#14654](https://github.com/elixir-lang/elixir/pull/14654) (which cites #198 as motivation), shards by topic key — better for workloads with many topics and few subscribers per topic, since key-based lookups touch a single partition. The default `group_by: :pid` (≡ `keys: {:duplicate, :pid}`) is still best when topics have many subscribers each.

See [`Registry.start_link/1`](https://hexdocs.pm/elixir/Registry.html#start_link/1) for trade-offs.

## Notes

- Using `group_by: :key` requires Elixir 1.19+

### DIFF
--- a/lib/phoenix/pubsub.ex
+++ b/lib/phoenix/pubsub.ex
@@ -174,6 +174,14 @@ defmodule Phoenix.PubSub do
     * `:dispatcher` - the default dispatcher module for broadcasts
       (defaults to `Phoenix.PubSub`). Can be overridden per-call by
       passing a dispatcher to `broadcast/4` and friends.
+    * `:group_by` - controls how the underlying `Registry` partitions
+      subscriptions, either `:pid` or `:key` (defaults to `:pid`). With
+      `:pid`, entries are grouped by subscriber pid — best when topics
+      have many subscribers each. With `:key`, entries are grouped by
+      topic so key-based lookups touch a single partition — best when
+      there are many topics with few subscribers each. `:key` requires
+      Elixir v1.19 or later. See `Registry.start_link/1` for the
+      underlying trade-offs.
 
   """
   @spec child_spec(keyword) :: Supervisor.child_spec()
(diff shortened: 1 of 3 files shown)


lib/phoenix/pubsub.ex
```diff
@@ -174,6 +174,18 @@ defmodule Phoenix.PubSub do
     * `:dispatcher` - the default dispatcher module for broadcasts
       (defaults to `Phoenix.PubSub`). Can be overridden per-call by
       passing a dispatcher to `broadcast/4` and friends.
+    * `:keys` - the key registration mode for the underlying `Registry`,
+      one of `:unique`, `:duplicate`, `{:duplicate, :pid}`, or
+      `{:duplicate, :key}` (defaults to `:duplicate`, which is equivalent
+      to `{:duplicate, :pid}`). With `:unique`, subscribing the same
+      process to the same topic twice returns
```
josevalim: I thought unique would only allow a single process per topic, no? It probably doesn't make sense? Perhaps the option should be `group_by: :pid | :key`?

studzien: Yes, that's right. It doesn't make much sense, but I can think of scenarios where it does (like limiting to one subscriber per user topic).
The main reason I went with just passing the option directly to `Registry` was that we can't use the `{:duplicate, _}` form in Elixir < 1.19 (it crashes in CI runs).
Would making an Elixir-dependent test be the way to go here?

josevalim: Default to :duplicate and use the tuple only of group_by is given. Let’s not support unique :)

studzien: Changed. Also added a test dependent on the Elixir version, but maybe we don't need that

josevalim:

PR #213, opened 2026-07-13
Fix stale full-state transfers overwriting newer values in Tracker.State

### LINKED ISSUE #148: Race condition causing data inconsistency when nodes are coming up
I think I found a race condition that is causing invalid data.

**Version**: latest master (7893228).

**Some background**: We use a slightly modified version of phoenix_pubsub which has some performance optimizations (one PR up here as well, + we added tag lookup ets table to speed up delta merges). We encountered a shard crash that seemed to happen because of our modifications. After I was able to create a failing test, I also noticed that the bug exists also on the original branch (though, in the original branch, it does not cause a shard crash but data inconsistency instead).

For us, this seems to happen when there's a network partition or Kubernetes thinks it's a good idea to move/add some pods around.

This is really hard to replicate in the real world. It usually happens for us maybe once a month.

**Scenario (same as in the test but in the words)**:
* Node A and B are connected
* Alice joins node A (meta = initial)
* Node A syncs with Node B
* Alice is changed (meta = update1) - this update has been sent out but has not reached Node B yet
* Node C joins - connects with A & B
* Alice is changed (meta = update2)
* Node C sends out transfer reqs (no response yet)
* Node C receives delta update from A (with latest alice)
* Node C receives transfer ack from Node B (which does not have the latest alice)
* **problem**: the latest alice is overwritten by the old alice from Node B

**Failing test**:
```elixir
test "delta before transfer from a different node", config do
  a = new(:a, config)
  b = new(:b, config)
  {a, _, _} = State.replica_up(a, b.replica)
  {b, _, _} = State.replica_up(b, a.replica)

  alice = new_pid()

  # Alice joins Node A
  a = State.join(a, alice, "lobby", :alice, "initial")

  # Node A sends updates to node B
  assert {b, [{{_, _, :alice}, _, _}], _} = State.merge(b, State.extract(a, b.replica, b.context))
  assert [:alice] = b |> State.online_list() |> keys()
  a = State.reset_delta(a)

  # Adam is updated first time
  a = State.leave(a, alice, "lobby", :alice)
  a = State.join(a, alice, "lobby", :alice, "update1")

  # update1 is not received by Node B (because of network delay or network
  # partition) or is received a lot later
  a = State.reset_delta(a)

  # Node C comes up
  c = new(:c, config)
  {b, _, _} = State.replica_up(b, c.replica)
  {a, _, _} = State.replica_up(a, c.replica)
  {c, _, _} = State.replica_up(c, a.replica)
  {c, _, _} = State.replica_up(c, b.replica)

  # Adam is updated second time
  a = State.leave(a, alice, "lobby", :alice)
  a = State.join(a, alice, "lobby", :alice, "second")

  # Lets assume Node C also sent out transfer_req to Node B here, but Node C
  # receives delta heartbeat from Node A first.
  assert {c, [{{_, _, :alice}, "second", _}], []} = State.merge(c, a.delta)

  # Here everything is fine. Node C sees the latest alice.
  assert [
    {{"lobby", _, :alice}, "second", _}
  ] = c |> State.online_list()

  # Now Node C receives transfer ack from B (who has alice with one missed update)
  assert {c, _, _} = State.merge(c, State.extract(b, c.replica, c.context))
  assert [
    {{"lobby", _, :alice}, "second", {{:a, 1}, 2}}
  ] = c |> State.online_list()
  # ^ This fails because the most recent alice is overwritten with the old
  # alice (who has "initial" now in the meta")

  # Lets say we ignore the previous inconsistency and wait for transfer ack
  # from the node A as well
  assert {c, _, _} = State.merge(c, State.extract(a, c.replica, c.context))
  assert [
    {{"lobby", _, :alice}, "second", _}
  ] = c |> State.online_list()
  # ^ This still fails - now there is no alice online at all
end
```
Also link: https://github.com/salemove/phoenix_pubsub/commit/4be0251d4ebeb3904a9e8fcaefcdb9d91ed55be5

**Note**: As this is quite complex to replicate in the real world, I cannot be 100% sure that my test is exactly what is happening. I'm fairly certain there's "values" overwriting happening because I was able to change this [line](https://github.com/phoenixframework/phoenix_pubsub/blob/7893228b48752437dff20b269ffdf614e07388dc/lib/phoenix/tracker/state.ex#L294)  to use `true = :ets.insert_new` and this threw an error when there were new pods coming up (it took 2 weeks to catch that though).

In case my assumptions and the test case are correct - I still don't have a good idea how to fix it...

davydog187:
Fixes #148

> [!WARNING]
> I used Fable to investigate this bug, cross reference the fix in our application, and create this patch

We have been seeing the following bug in our application more recently, particularly around deployments when our devices reconnect to our platform.

## The bug

#148 reports a race when nodes come up mid-churn: a stale full-state `transfer_ack` from a partitioned replica overwrites a newer value the receiving node already holds. We hit the downstream symptom of this in production (~245 shard crashes over a month on a Fly.io cluster with rolling deploys): `** (stop) {:badmatch, 3}` in `Phoenix.Tracker.State.remove/4`.

The mechanism, following #148's scenario (A, B connected; alice joins A; A syncs to B; alice updated twice on A but B misses the updates; fresh node C receives A's delta first, then B's stale transfer):

1. C merges A's *delta* carrying alice's newest dot (e.g. `{a, 5}`). Because C never observed dots 1–4, the dot parks in C's cloud and C's context for `a` stays at 0 — a legitimate compaction gap.
2. B's stale full-state extract arrives carrying old alice with dot `{a, 1}`. In `accumulate_joins`, admission is decided only by `in?(context, clouds, tag)` — dot 1 is neither in C's context (0) nor its cloud (`{3,4,5}`) — so the stale element is **admitted**, even though C holds the same `{topic, pid, key}` with a strictly newer dot from the same replica. `observe_removes` doesn't flag C's newer row either (B never observed dot 5).
3. The two ETS tables then diverge: the `values` **ordered_set** insert *overwrites* the newer row (metadata downgraded; after a subsequent merge of A's full state the presence can disappear entirely), while the `pids` **duplicate_bag** *appends* a duplicate `{pid, topic, key}` row. Each stale merge appends another.
4. A later local leave hits `1 = :ets.select_delete(pids, ...)` in `remove/4` → `{:badmatch, N}` → shard crash.

## The fix

Two complementary changes in `State`:

- **`accumulate_joins` per-key dominance check**: skip a remote element when the local set already holds the same `{topic, pid, key}` with a same-replica dot `>=` the incoming one. A stale dot can never displace a newer element from its own origin replica, so no resurrection and no metadata downgrade. Genuinely concurrent adds (different origin replica, or a newer dot) are unaffected — the context/cloud machinery tracks *observed dots*, and this adds the missing notion of *per-key dominance* without touching add-wins semantics.
- **`merge/3` pids consistency guard**: don't insert a `pids` entry for a join whose `values` row already exists — that's an in-place update and the pids row is already present. This makes it structurally impossible for `values`/`pids` to diverge even if a conflicting tag is admitted through some other path (e.g. a replica reincarnation window before permdown).

Relaxing the `1 =` assertions in `remove/4` was considered and rejected: it would only mask ongoing state corruption that the assertion currently surfaces.

## Tests

All three fail on `main` with the genuine bug symptoms and pass with the fix:

- `state_test.exs`: the #148 scenario — alice's meta must stay `"second"` after merging B's stale extract, no duplicate pids rows, and alice must survive a subsequent merge of A's full state (on `main` she reverts to `"initial"` and then disappears).
- `state_test.exs`: the production crash — two independent stale transfers accumulate duplicate pids rows and `State.leave/2` raises `MatchError` (the `{:badmatch, N}`).
- `shard_race_test.exs` (new): end-to-end on real `:peer` nodes with real shards. Real replication builds the preconditions (B partitioned via `drop_gossips`, C boots mid-stream with the compaction gap — asserted explicitly before the bug assertion); the decisive stale `transfer_ack` is built from B's genuinely-stale state and injected for determinism, since transfer target selection isn't controllable from a test. Asserts the raw CRDT row keeps the newest meta and tag, exactly one pids row, alice online end-to-end, and the shard survives a subsequent leave.

Full suite is green across repeated runs (`129 tests, 0 failures`).

### DIFF
--- a/lib/phoenix/tracker/state.ex
+++ b/lib/phoenix/tracker/state.ex
@@ -348,7 +348,15 @@ defmodule Phoenix.Tracker.State do
     value_keys_to_remove = removed_value_keys -- added_value_keys
 
     pids_to_remove = removed_pids -- added_pids
-    pids_to_add = added_pids -- removed_pids
+
+    # A join whose value key is already present is an in-place update:
+    # the values insert below overwrites the row, so the pids entry
+    # must not be inserted again or the duplicate_bag accumulates
+    # duplicates that crash a later remove/4.
+    pids_to_add =
+      for {pid, topic, key} <- Enum.uniq(added_pids -- removed_pids),
+          not :ets.member(local.values, {topic, pid, key}),
+          do: {pid, topic, key}
 
     for value_key <- value_keys_to_remove do
       :ets.delete(local.values, value_key)
@@ -374,16 +382,33 @@ defmodule Phoenix.Tracker.State do
 
   @spec accumulate_joins(t, values) :: joins :: {[pid_lookup], [values]}
   defp accumulate_joins(local, remote_map) do
-    %State{context: context, clouds: clouds} = local
+    %State{context: context, clouds: clouds, values: values} = local
 
     Enum.reduce(remote_map, {[], []}, fn {{replica, _} = tag, {pid, topic, key, meta}},
                                          {pids, adds} ->
-      if not match?(%{^replica => _}, context) or in?(context, clouds, tag) do
-        {pids, adds}
-      else
-        {[{pid, topic, key} | pids], [{{topic, pid, key}, meta, tag} | adds]}
+      cond do
+        not match?(%{^replica => _}, context) or in?(context, clouds, tag) ->
+          {pids, adds}
+
+        superseded_locally?(values, pid, topic, key, tag) ->
+          {pids, adds}
+
+        true ->
+          {[{pid, topic, key} | pids], [{{topic, pid, key}, meta, tag} | adds]}
       end
     end)
+  end
+
+  # A remote element is stale if we already hold the same {topic, pid, key}
+  # with a newer dot from the same origin replica. This happens when a
+  # replica's full-state transfer races a delta we already applied: our
+  # context has a gap below our dot, so the old dot is not covered by
+  # `in?/3`, but it must not overwrite the newer element we hold.
+  defp superseded_locally?(values, pid, topic, key, {replica, clock}) do
+    case :ets.lookup(values, {topic, pid, key}) do
+      [{_, _meta, {^replica, local_clock}}] -> local_clock >= clock
+      _ -> false
+    end
   end
 
   @spec observe_removes(t, t, map) ::
--- a/lib/phoenix/tracker/state.ex
+++ b/lib/phoenix/tracker/state.ex
@@ -404,6 +404,14 @@ defmodule Phoenix.Tracker.State do
   # replica's full-state transfer races a delta we already applied: our
   # context has a gap below our dot, so the old dot is not covered by
   # `in?/3`, but it must not overwrite the newer element we hold.
+  #
+  # This is sound because we only compare dots from the *same* origin replica
+  # for the *same* {topic, pid, key}. Tracker's invariant is one live element
+  # per key per replica with monotonically increasing dots, so a lower
+  # same-replica dot is provably a superseded add; skipping it cannot drop a
+  # genuinely concurrent add (a different replica, or a higher dot, both fall
+  # through to the `true` branch). This is a deliberate deviation from a
+  # textbook ORSWOT merge, which has no per-replica dominance shortcut.
   defp superseded_locally?(values, pid, topic, key, {replica, clock}) do
     case :ets.lookup(values, {topic, pid, key}) do
       [{_, _meta, {^replica, local_clock}}] -> local_clock >= clock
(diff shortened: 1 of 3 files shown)


lib/phoenix/tracker/state.ex
```diff
@@ -348,7 +348,15 @@ defmodule Phoenix.Tracker.State do
...
-    pids_to_add = added_pids -- removed_pids
+
+    # A join whose value key is already present is an in-place update:
+    # the values insert below overwrites the row, so the pids entry
+    # must not be inserted again or the duplicate_bag accumulates
+    # duplicates that crash a later remove/4.
+    pids_to_add =
+      for {pid, topic, key} <- Enum.uniq(added_pids -- removed_pids),
+          not :ets.member(local.values, {topic, pid, key}),
+          do: {pid, topic, key}
```
davydog187, line 359: Is there a more efficient way to do this? Does this process own writing to the ets table, do we need to worry about other writers?

davydog187: Ok, the Shard is the sole writer. This should be safe and reasonably efficient.

PR #217, opened 2026-08-25
Add tagged subscriptions

SteffenDE:
This is meant for situations where one process subscribes to the same topic multiple times independently and wants to differentiate the different subscriptions - or - a process subscribes to two different topics that share the same message format and wants to differentiate the topics.

In this case, it can now add a tag when subscribing and messages will arrive as `{tag, message}` instead of `message`, assuming the default dispatcher is used or the custom dispatcher handles tagging.

### DIFF
--- a/lib/phoenix/pubsub.ex
+++ b/lib/phoenix/pubsub.ex
@@ -60,6 +60,43 @@ defmodule Phoenix.PubSub do
   to thousands or even millions of users to be encoded once and written
   directly to sockets instead of being encoded per channel.
 
+  ## Tagged subscriptions
+
+  A process may subscribe to the same topic more than once. By default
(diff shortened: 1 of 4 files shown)


lib/phoenix/pubsub.ex
```diff
@@ -396,20 +497,63 @@ defmodule Phoenix.PubSub do
     adapter.node_name(name)
   end
 
+  @doc """
+  Applies a subscription's tag to `message`.
```
josevalim, line 501: Do we really need this? Custom dispatchers could just implement their own tags themselves.

SteffenDE:
Closing in favor of https://github.com/phoenixframework/phoenix_pubsub/pull/218.

PR #218, opened 2026-09-11
Message senders

SteffenDE:


lib/phoenix/pubsub.ex
```diff
@@ -213,7 +218,13 @@ defmodule Phoenix.PubSub do
   @spec subscribe(t, topic, keyword) :: :ok | {:error, term}
   def subscribe(pubsub, topic, opts \\ [])
       when is_atom(pubsub) and is_binary(topic) and is_list(opts) do
-    case Registry.register(pubsub, topic, opts[:metadata]) do
+    meta =
+      cond do
+        opts[:sender] -> {__MODULE__, :__sender__, opts[:sender]}
+        true -> opts[:metadata]
+      end
```
josevalim, line 225: This is going to ballon the amount of data. My suggestion is to do:


```suggestion
    meta =
      case opts[:sender] do
        {mod, term} when is_atom(mod) -> {mod, term}
        _ ->
          # TODO: Deprecate me
          case opts[:metadata] do
            {mod, _} when is_atom(mod) -> warn; other
            nil -> nil
            other -> other
          end
      end
```

Basically, we should keep the sender path as efficient as possible. And in case of conflict, we raise/warn. It should be very unlikely anyway.

lib/phoenix/pubsub/pg2.ex
```diff
@@ -116,7 +116,7 @@ defmodule Phoenix.PubSub.PG2Worker do
 
   @impl true
   def handle_info({:forward_to_local, topic, message, dispatcher}, pubsub) do
-    Phoenix.PubSub.local_broadcast(pubsub, topic, message, dispatcher)
+    Phoenix.PubSub.local_dispatch(pubsub, topic, message, dispatcher)
```
josevalim, line 119: This means other adapters will have the same issue, right? They will need a way to call `local_broadcast` without emitting a deprecation warning. I wonder if we postpone the deprecation of the dispatcher in `local_broadcast` to the same time we deprecate `:metadata`...

SteffenDE: The question is when to deprecate with a compile time warning. With the potential metadata conflict, I feel like we should do 3.0 anyway. And then fully remove support in 4.0.

josevalim: We can deprecate in the last version before we ship v3.0 and then release v3.0 without it. But keeping a dispatch only on local_broadcast itself is not going to hurt, so agreed with your timeline.

test/support/remote_dispatcher.ex
```diff
@@ -0,0 +1,24 @@
...
+
+defmodule Phoenix.PubSub.RemoteSender do
+  @moduledoc false
+  @behaviour Phoenix.PubSub.Sender
+
+  @impl true
+  def send(pid, meta, message, state) do
+    Kernel.send(pid, {:sent, node(), meta, message, state})
+    state
+  end
```
josevalim, line 23: We are not testing the state is being passed around I believe.

josevalim (approved):
Looks good to me! Let's make sure Phoenix v1.8 can run this without runtime warnings and it is good to go!

test/support/remote_dispatcher.ex
SteffenDE: we have separate tests for the state

SteffenDE: https://github.com/phoenixframework/phoenix_pubsub/pull/218/changes#diff-cb4cad8705e021b58b92bb56cf0fa648391471e1cd60dc1be2e650a1d4d4ed67R107-R116

